PR #28, opened 2020-03-18
Parts ut

lewisrenfrew:
- nowhere near finished but PRing now since getting big


src/Domain.LinnApps/Parts/Part.cs
```diff
@@ -4,7 +4,7 @@
 
     public class Part
     {
-        public int BridgeId { get; set; }
+        public long Id { get; set; }
```
richardiphillips: won't int be enough? I don't think we'll have a billion parts too soon

src/Domain.LinnApps/ParetoClass.cs
```diff
@@ -1,9 +1,15 @@
...
+
+    using Domain.LinnApps.Parts;
+
     public class ParetoClass
     {
         public string ParetoCode { get; set; }
 
         public string Description { get; set; }
+
+        public List<Part> Parts { get; set; }
```
richardiphillips: does this have to be list.

src/Resources/PartResource.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public long Id { get; set; }
+
+        public string PartNumber { get; set; }
+
+        public string Description { get; set; }
+
+        public string RootProduct { get; set; }
+
+        public string StockControlled { get; set; }
```
richardiphillips: may be better as bools?

src/Domain.LinnApps/Parts/Part.cs
lewisrenfrew: int is definitely enough! I was playing around with hashing part numbers to make aId's (which ended up being very big numbers) and forgot to revert back to int. Will do now.

PR #31, opened 2020-03-23
Parts ut 3

lewisrenfrew:


src/Domain.LinnApps/SernosSequence.cs
```diff
@@ -5,7 +5,5 @@ public class SernosSequence
         public string Sequence { get; set; }
 
         public string Description { get; set; }
-
-        public string DecrementRule { get; set; }
```
lewisrenfrew: I don't remember adding this field and it didn't seem used so I took it out. Any objections let me know

src/Service.Host/index.html
```diff
@@ -10,7 +10,7 @@
     <script type="text/javascript">
         window.APPLICATION_SETTINGS = {
             authorityUri: 'https://www-sys.linn.co.uk/auth/',
-            appRoot: 'http://localhost:61798',
+            appRoot: 'http://localhost:51698',
```
iaincrawf: Do we just pick a number for these that's different from the rest of the solutions?

PR #32, opened 2020-03-26
Partial parcel ut back-end

iaincrawf:


src/Facade/Services/IFacadeWithSearchReturnTen.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Linn.Stores.Facade.Services
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+
+    public interface IFacadeWithSearchReturnTen<T, in TKey, in TResource, in TUpdateResource> : IFacadeService<T, TKey, TResource, TUpdateResource>
```
lewisrenfrew: Does this need to be so specific? How about a search method with an option limit parameter that you could set to ten, or anything you like. I'd also include the word 'Service' in the interface name somewhere for conventions sake.

src/Facade/Services/SuppliersService.cs
```diff
@@ -1,24 +1,77 @@
-﻿namespace Linn.Stores.Facade.Services
+﻿//namespace Linn.Stores.Facade.Services
```
lewisrenfrew: not sure what's going on here but I wouldn't let this commented out code into master

iaincrawf: Oops yeah I'll take that out, wasn't going to pr it like that last night then put it up today without checking

src/Facade/Services/IFacadeWithSearchReturnTen.cs
iaincrawf: Yeah I can change it to have an optional limit and chuck service in the name, I basically just copied it in from production to use it in a similar way

PR #33, opened 2020-03-26
Parts ut react client

lewisrenfrew:
- another wip PR, but I'm off for a few days and don't wanna leave this hanging


src/IoC/PersistenceModule.cs
```diff
@@ -23,7 +23,7 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<PartRepository>().As<IRepository<Part, int>>();
             builder.RegisterType<ParetoClassRepository>().As<IRepository<ParetoClass, string>>();
             builder.RegisterType<DepartmentRepository>().As<IQueryRepository<Department>>();
-            builder.RegisterType<ProductAnalysisCodeRepository>().As<IRepository<ProductAnalysisCode, string>>();
+            builder.RegisterType<ProductAnalysisCodeRepository>().As<IQueryRepository<ProductAnalysisCode>>();
```
richardiphillips: why do you use query repo?

lewisrenfrew: Just because I didn't need to do any updating or adding or anything and it meant less unimplemented methods. I can change it back if we will be creating and editing ProductAnalysisCodes in this solution. Will come back to it in my next pr

PR #35, opened 2020-03-30
Alloc options

richardiphillips:
Intermediate PR since I'm going to be on support now


richardiphillips:
> 👍 is the options to allocate by country and sales accounts coming later?

Yep. and the other boolean options etc. Will probably try and get accounts from sales accounts repo

PR #36, opened 2020-04-01
Parts ut front end 2

lewisrenfrew:
- general and build tabs only


src/Service.Host/client/src/components/parts/tabs/PurchTab.js
```diff
@@ -0,0 +1,337 @@
...
+    oneOffRequirement,
+    sparesRequirement,
+    ignoreWorkstationStock,
+    handleIgnoreWorkstationStockChange,
+    imdsIdNumber,
+    imdsWeight,
+    mechanicalOrElectronic,
+    partCategories
+}) {
+    const convertToYOrNString = booleanValue => {
```
iaincrawf, line 35: Not sure if it's worth it but if this is wee func is used in a enough places would there be a benefit to making it somewhere centralised/its own component?

tests/Integration/Service.Tests/AccountingCompaniesModuleSpecs/WhenGettingAll.cs
```diff
@@ -27,7 +27,7 @@ public void SetUp()
                                };
             var companyB = new AccountingCompany
                                {
-                                   Name = "B", Description = "description B"
+                                   Name = "B", Description = "description B yeah I think this would be doable for work stuff tbh"
```
iaincrawf, line 30: 10/10 for test description haha

iaincrawf (approved):
Looks good to me!

src/Facade/Services/PartFacadeService.cs
```diff
@@ -69,6 +106,8 @@ protected override void UpdateFromResource(Part entity, PartResource resource)
             entity.CccCriticalPart = this.ToYesOrNoString(resource.CccCriticalPart);
             entity.EmcCriticalPart = this.ToYesOrNoString(resource.EmcCriticalPart);
             entity.SafetyCriticalPart = this.ToYesOrNoString(resource.SafetyCriticalPart);
+            entity.PlannedSurplus = this.ToYesOrNoString(resource.PlannedSurplus);
```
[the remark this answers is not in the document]
iaincrawf: Saw the js version later on so guessing we made it and it does the same, nevermind lol

src/Service.Host/client/src/components/parts/tabs/GeneralTab.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
```
iaincrawf, line 1: Are there any tests for these components btw?

lewisrenfrew: none at all.. They're pretty massive components and I can't really think of any useful tests. They're pretty much exclusively made up of shared components though which should be tested (in theory) so it's not entirely untested code.

PR #41, opened 2020-09-30
Part triggers

lewisrenfrew:


src/Domain.LinnApps/Parts/PartService.cs
```diff
@@ -113,10 +124,84 @@ public Part CreatePart(Part partToCreate, List<string> privileges)
         {
             if (!this.authService.HasPermissionFor(AuthorisedAction.PartAdmin, privileges))
```
atcampbell, line 125: Same for permission checks?

src/Domain.LinnApps/Parts/PartService.cs
```diff
@@ -1,18 +1,46 @@
...
+
+        private readonly IRepository<QcControl, int> qcControlRepository;
+
+        private readonly IRepository<Part, int> partRepository;
+
+        private readonly IRepository<PartTemplate, string> templateRepository;
+
+        private readonly IPartPack partPack;
+
+        private readonly ITransactionManager transactionManager;
```
atcampbell, line 26: Should a lot of this be in a facade service rather than being in the domain?

src/Persistence.LinnApps/Repositories/PartRepository.cs
```diff
@@ -42,7 +42,7 @@ public Part FindById(int key)
 
         public IQueryable<Part> FindAll()
         {
-            return this.serviceDbContext.Parts;
+            return this.serviceDbContext.Parts.AsNoTracking();
```
atcampbell, line 45: What does asNoTracking do again?

atcampbell (approved):

richardiphillips (approved):

src/Persistence.LinnApps/Repositories/PartRepository.cs
richardiphillips: doesn't cache anything so you can't update them but takes up less memory  and time

src/Domain.LinnApps/Parts/PartService.cs
richardiphillips: I'd always put as much as possible in the domain but the transaction manager should maybe be the level above. Was this to do with the post-commit triggers?

src/Persistence.LinnApps/Repositories/PartRepository.cs
richardiphillips: we should maybe/probably do this by default for getalls

src/Domain.LinnApps/Parts/PartService.cs
lewisrenfrew: I'm happy with everything being in the domain except that transaction manager. To be honest it might not even be necessary. My thinking was that all I've done in this file is update the PartTemplate object that corresponds to the table. Changes to that object won't be persisted until a Commit() is called, no? A Commit() might already be happening further downstream that catches this anyway so I'll double-check. Anyway, I should definitely move it out of the domain if I do need to keep it in.

lewisrenfrew: I think I am getting Commit() confused with SaveChanges()

PR #52, opened 2020-11-04
try docker timezone

lewisrenfrew:


atcampbell (approved):
In Production I had to set this to `ENV TZ=Europe/London` as the linn week pack was messing up with BST when the env was set to UTC. I need to look into that further but might be safer putting that here.

PR #53, opened 2020-11-05
Parcel ut backend

iaincrawf:
(hopefully complete backend)


src/Persistence.LinnApps/Repositories/ParcelRepository.cs
```diff
@@ -0,0 +1,53 @@
...
+            throw new NotImplementedException();
+        }
+
+        public IQueryable<Parcel> FilterBy(Expression<Func<Parcel, bool>> expression)
+        {
+            return this.serviceDbContext
+                .Parcels
+                .Where(expression)
+                .ToList()
+                .AsQueryable();
```
atcampbell, line 50: I think these last two might be redundant?

src/Service/Modules/ParcelsModule.cs
```diff
@@ -23,37 +23,42 @@ public ParcelsModule(IFacadeService<Parcel, int, ParcelResource, ParcelResource>
...
+
             var results = string.IsNullOrEmpty(resource.SearchTerm)
                               ? this.parcelsFacadeService.GetAll()
                               : this.parcelsFacadeService.Search(resource.SearchTerm);
+
             return this.Negotiate
                 .WithModel(results)
-                .WithMediaRangeModel("text/html", ApplicationSettings.Get)
-                .WithView("Index");
+                .WithMediaRangeModel("text/html", ApplicationSettings.Get);
```
atcampbell, line 44: I might be totally off here, but is the WithView needed here to serve up the app if going directly to the url in the browser?

atcampbell (approved):

src/Service/Modules/ParcelsModule.cs
iaincrawf: I think it would be needed but I've got a bit in the constructor that's been cut off by github which I think should do it! The search will be after page load so I think I want the intial pageload call not to bother doing a getAll or search -
` this.Get("/logistics/parcels", _ => this.Negotiate.WithModel(ApplicationSettings.Get()).WithView("Index"));`

PR #73, opened 2020-12-22
Mech part source purchasing tabs

lewisrenfrew:


src/Domain.LinnApps/Parts/MechPartSource.cs
```diff
@@ -161,5 +161,55 @@ public class MechPartSource
         public string FootprintRef { get; set; }
 
         public string RkmCode { get; set; }
+
+        public ICollection<MechPartPurchasingQuote> PurchasingQuotes { get; set; }
```
atcampbell, line 165: Why are you using ICollection instead of IEnumerable?

src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -90,10 +90,15 @@ public class ServiceDbContext : DbContext
 
         public DbSet<PartParamData> PartParamDataSheets { get; set; }
 
+        public DbSet<MechPartPurchasingQuote> MechPartPurchasingQuotes { get; set; }
+
```
atcampbell, line 94: Some extra lines

src/Facade/ResourceBuilders/MechPartSourceResourceBuilder.cs
```diff
@@ -101,8 +106,47 @@ public MechPartSourceResource Build(MechPartSource model)
...
+                            RemoveTCodeByName = model.RemoveTCodeBy?.FullName,
+                            RemoveTCodeDate = model.RemoveTCodeDate?.ToString("o"),
+                            CancelledBy = model.CancelledById,
+                            CancelledByName = model.CancelledBy?.FullName,
+                            CancelledDate = model.DateCancelled?.ToString("o"),
+                            PurchasingQuotes = model.PurchasingQuotes?.Select(
+                                (q, i) =>
+                                    {
+                                        var resource = this.purchasingQuotesResourceBuilder.Build(q);
+                                        resource.Id = i;
```
atcampbell, line 135: Do these ids need to be the iterator?

src/Resources/Parts/MechPartSourceResource.cs
```diff
@@ -166,8 +166,58 @@ public class MechPartSourceResource : HypermediaResource
...
+
+        public string RemoveTCodeDate { get; set; }
+
+        public int? CancelledBy { get; set; }
+
+        public string CancelledByName { get; set; }
+
+        public string CancelledDate { get; set; }
+
+        public List<MechPartPurchasingQuoteResource> PurchasingQuotes { get; set; }
```
atcampbell, line 215: Think these should be IEnumerables

atcampbell (approved):

src/Domain.LinnApps/Parts/MechPartSource.cs
lewisrenfrew: I need to add, remove and update elements in the these lists. I don't think IEnumerable lets you do that? Pretty sure I got errors when these were IEnumerables so just switched to ICollection

src/Facade/ResourceBuilders/MechPartSourceResourceBuilder.cs
lewisrenfrew: Yeah, I think so. The underlying entity Id is supplierId or something, but since you can change these in the table it doesn't make for a very reliable Id for react map keys.

atcampbell: Is this better done in the client if the fields are only used by React? It seems weird to put this in the resource when its not really a field on the object (& just used in a react array)

PR #75, opened 2021-01-07
update component library for editaable table fix

atcampbell:


iaincrawf (approved):
Reverting sounds good to me if there's something broken in live, especially if you've got some more improvements you fancy doing

iaincrawf:
> Reverting sounds good to me if there's something broken in live, especially if you've got some more improvements you fancy doing

Hahaha sorry I've got mixed up and managed to read this wrongly, a fix is even better!!

PR #78, opened 2021-01-20
Dept stock ut

lewisrenfrew:
- waiting on the SingleEditableTable for the client, this won't pass it's checks until https://github.com/linn/react-components-library/pull/201 is merged
- also need to make sure anything using the old table is updated before this is merged


src/Domain.LinnApps/Parts/PartService.cs
```diff
@@ -215,6 +220,27 @@ public Part CreateFromSource(int sourceId, int createdBy)
...
+                        ||
+                        ((p.StockControlled.Equals("N") 
+                          && (p.BaseUnitPrice == 0 || p.BaseUnitPrice == null)
+                          && (p.LinnProduced.Equals("N") || p.LinnProduced == null))
+                        &&
+                        !(from s in stockLocators
+                         where s.PartNumber == p.PartNumber
+                         select s.PartNumber)
+                        .Contains(p.PartNumber))
+                    select p;
```
atcampbell, line 241: Could this be split up to make it a bit clearer? Just thinking it might be hard to follow / debug

src/Domain.LinnApps/StockLocators/StockLocatorService.cs
```diff
@@ -0,0 +1,122 @@
...
+            IRepository<StockLocator, int> repository,
+            IRepository<StoresPallet, int> palletRepository,
+            IQueryRepository<StoragePlace> storagePlaceRepository)
+        {
+            this.repository = repository;
+            this.palletRepository = palletRepository;
+            this.storagePlaceRepository = storagePlaceRepository;
+        }
+
+        public void UpdateStockLocator(StockLocator @from, StockLocator to)
```
atcampbell, line 27: Is this @ a typo?

src/Domain.LinnApps/StockLocators/StockLocatorService.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            var stockLocators = this.repository
+                .FilterBy(s => s.PartNumber == partNumber);
+
+            string auditDept = string.Empty;
+
+            return stockLocators.ToList().Select(
+                s =>
+                    {
+                        var l = this.storagePlaceRepository.FindBy(
```
atcampbell, line 96: Maybe some of these var names could be a bit clearer

src/Facade/ResourceBuilders/StockLocatorResourceBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    using Linn.Common.Facade;
+    using Linn.Stores.Domain.LinnApps.StockLocators;
+    using Linn.Stores.Resources;
+
+    public class StockLocatorResourceBuilder : IResourceBuilder<StockLocator>
+    {
+        public StockLocatorResource Build(StockLocator stockLocator)
+        {
+            if (stockLocator == null)
```
atcampbell, line 13: Does this ever get called with a null stock locator?

atcampbell (approved):

src/Domain.LinnApps/StockLocators/StockLocatorService.cs
lewisrenfrew: No, I think from is a reserved word and so the @ is needed to escape it? Not sure. Could just rename the parameter.

src/Facade/ResourceBuilders/StockLocatorResourceBuilder.cs
lewisrenfrew: Not anymore. I was returning a null response in the delete case initially, but I changed it to return the deleted item. Good spot

PR #88, opened 2021-02-03
Feedback from despatch picking summary report. 

richardiphillips:
Added and imported a landscape print style for now until we decide on a better way...

### DIFF
--- a/src/Service.Host/client/src/css/landscapePrint.css
+++ b/src/Service.Host/client/src/css/landscapePrint.css
@@ -1,0 +1,16 @@
+@media print {
+    @page {
+        size: 297mm 210mm ; 
+        
+        margin: 0mm;
+        margin-right: 0mm; 
+      }
+}
+
+.date-for-printing {
+    display: none;
+    right: 40px;
+    top: 40px;
+    position: absolute;
+    font-size: small;
+}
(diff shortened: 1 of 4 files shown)


iaincrawf (approved):

src/Service.Host/client/src/css/landscapePrint.css
```diff
@@ -0,0 +1,16 @@
...
+    @page {
+        size: 297mm 210mm ; 
+        
+        margin: 0mm;
+        margin-right: 0mm; 
+      }
+}
+
+.date-for-printing {
+    display: none;
```
iaincrawf, line 11: Does the date show in the print preview? Could be wrong, or maybe it uses the shared css and this just overrides the sizes, but I thought it'd need a display: block inside the @media print bit too. Probably doesn't matter if I'm hopefully going to put it into the shared css in a mo anyway

iaincrawf: Ah it'll use the shared one yeah! Ignore that ^

richardiphillips: yes. it shows when printing. it still picks that up from the shared @media print. I've only overridden the size

PR #93, opened 2021-02-09
Import Book domain classes & resources 

atcampbell:


src/Facade/Services/ParcelFacadeService.cs
```diff
@@ -69,4 +69,6 @@ protected override void UpdateFromResource(Parcel entity, ParcelResource updateR
             throw new NotImplementedException();
         }
     }
+
+    // public class ImportBookFacadeService : IFacadeService<>
```
iaincrawf, line 73: wee comment left in but it'll get overwritten whenever I get parcel stuff PR'd anyway

iaincrawf (approved):
looks good to me

PR #95, opened 2021-02-11
Parcel ut partial front end 

iaincrawf:
Have kept it out of the menu for now

Search parcels is implemented from front end, still need to finish full query in service as currently on searching on parcel number

Still need to do add/edit parcel page


src/Service.Host/client/src/components/parcels/ParcelsSearch.js
```diff
@@ -0,0 +1,329 @@
...
+                <Grid item xs={2}>
+                    <SearchInputField
+                        label="Date Created"
+                        fullWidth
+                        placeholder="search.."
+                        onChange={handleDateSearchChange}
+                        propertyName="dateCreatedSearchTerm"
+                        type="date"
+                        value={dateCreatedSearch}
+                        debounce={5000}
```
lewisrenfrew, line 235: long debounce?

src/Service.Host/client/src/components/parcels/ParcelsSearch.js
```diff
@@ -0,0 +1,329 @@
...
+        setSearchTerm(
+            `${parcelNumberSearch}&supplierIdSearchTerm=${supplierIdSearch}&carrierIdSearchTerm=${carrierIdSearch}&dateCreatedSearchTerm=${dateCreatedSearch}&supplierInvNoSearchTerm=${supplierInvNoSearch}&consignmentNoSearchTerm=${args[1]}&commentsSearchTerm=${commentsSearch}`
+        );
+    };
+
+    const handleCommentSearchChange = (...args) => {
+        setCommentsSearch(args[1]);
+        setSearchTerm(
+            `${parcelNumberSearch}&supplierIdSearchTerm=${supplierIdSearch}&carrierIdSearchTerm=${carrierIdSearch}&dateCreatedSearchTerm=${dateCreatedSearch}&supplierInvNoSearchTerm=${supplierInvNoSearch}&consignmentNoSearchTerm=${consignmentNoSearch}&commentsSearchTerm=${args[1]}`
+        );
```
lewisrenfrew, line 93: These search strings are quite unwieldly but I'm not sure there's a better option..
You could store all the searchTerms in one state object

```
const [searchTerms, setSearchTerms] = useState({
    parcelNumberSearch: '',
    supplierIdSearch: '',
   etc... }
);
```

and then write a little function to build the big query strings by looping over the searchTerms keys and values, or something like that.

If it ends up being even messier then just leave it, but a little utility function (or hook?) to turn objects to searchTerms could be useful

lewisrenfrew: we also already have this package as a dependency that could help https://www.npmjs.com/package/query-string

lewisrenfrew (commented):
looks good. few commented out sections you could remove just for cleanliness but I imagine you'll be coming back to this soon anyway

src/Service.Host/client/src/components/parcels/ParcelsSearch.js
richardiphillips: +1 query-string

src/Service.Host/client/src/components/parcels/ParcelsSearch.js
iaincrawf: I don't think these even take a debounce param, will take it out

src/Service.Host/client/src/components/parcels/ParcelsSearch.js
iaincrawf: query string sounds good to me, I'll look it up properly and put it in

PR #96, opened 2021-02-15
Wand screen

richardiphillips:
Start of wand screen with alert to test scanner

### DIFF
--- a/src/Persistence.LinnApps/Repositories/WandConsignmentsRepository.cs
+++ b/src/Persistence.LinnApps/Repositories/WandConsignmentsRepository.cs
@@ -1,0 +1,34 @@
+﻿namespace Linn.Stores.Persistence.LinnApps.Repositories
+{
+    using System;
+    using System.Linq;
+    using System.Linq.Expressions;
+
+    using Linn.Common.Persistence;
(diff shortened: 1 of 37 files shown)


src/Persistence.LinnApps/Repositories/WandConsignmentsRepository.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public IQueryable<WandConsignment> FilterBy(Expression<Func<WandConsignment, bool>> expression)
+        {
+            throw new NotImplementedException();
+        }
+
+        IQueryable<WandConsignment> IQueryRepository<WandConsignment>.FindAll()
```
atcampbell, line 29: Should these be in the standard form of `public IQueryable<WandConsignment> FindAll()`

atcampbell (approved):

src/Persistence.LinnApps/Repositories/WandConsignmentsRepository.cs
richardiphillips: good spot. thanks I don't know how I managed that.

PR #97, opened 2021-02-16
Stock locator upd

lewisrenfrew:


src/Domain.LinnApps/StockLocators/StockLocatorBatchesViewModel.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Linn.Stores.Domain.LinnApps.StockLocators
+{
+    using System;
+
+    public class StockLocatorBatchesViewModel
```
richardiphillips, line 5: i used to call things model and I got argued out of it as it isn't descriptive of what the class represents. I compromise now by putting them in a models folder.

richardiphillips (approved):

PR #99, opened 2021-02-17
Part use reducer

lewisrenfrew:
example of useReducer to potentially simplify spaghetti state management for a big component.
- behaves exactly like a redux reducer 
- src/Service.Host/client/src/components/parts/Part.js is the file of interest


src/Service.Host/client/src/components/parts/Part.js
```diff
@@ -433,106 +462,94 @@ function Part({
...
                                     safetyCriticalHelperText={
                                         creating() ? getSafetyCriticalHelperText() : null
                                     }
-                                    performanceCriticalPart={part.performanceCriticalPart}
-                                    emcCriticalPart={part.emcCriticalPart}
-                                    singleSourcePart={part.singleSourcePart}
-                                    cccCriticalPart={part.cccCriticalPart}
-                                    psuPart={part.psuPart}
+                                    performanceCriticalPart={state.part.performanceCriticalPart}
+                                    emcCriticalPart={state.part.emcCriticalPart}
```
atcampbell, line 478: Could you destructure part from state somewhere / somehow?

atcampbell (approved):
I wonder about testing as well. I guess if its in the component it should be tested by the interactions rather than testing the reducer itself?

src/Service.Host/client/src/components/parts/Part.js
```diff
@@ -34,15 +34,206 @@ function Part({
...
+                    part: { ...state.part, [action.fieldName]: action.payload }
+                };
+            default:
+                return state;
+        }
+    }
+
+    // this useReducer call replaces all the old useState calls
+    // we now have one state object and one consistent method of updating it
+    const [state, dispatch] = useReducer(partReducer, {
```
atcampbell, line 183: Just thinking as well if it would be better just to have state as 'part'? prevPart could still be separate and might make things cleaner? Might also be better with both in the reducer...

src/Service.Host/client/src/components/parts/Part.js
```diff
@@ -34,15 +34,214 @@ function Part({
...
+        switch (action.type) {
+            case 'initialise':
+                if (creating()) {
+                    return { ...state, part: defaultPart };
+                }
+                return { ...state, part: action.payload, prevPart: action.payload };
+            case 'loadTemplate':
+                return { ...state, part: action.payload };
+            case 'fieldChange':
+                if (action.fieldName === 'rawOrFinished') {
```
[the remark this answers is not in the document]
lewisrenfrew: Yeah,

> I wonder about testing as well. I guess if its in the component it should be tested by the interactions rather than testing the reducer itself?

I guess so. I thought about trying to extract the reducer to a separate file. But it relies on having access to props which makes that tricky. I actually wonder whether relying on external things is allowed in the world of pure reducer functions now that I think about it more. I guess I could pass any prop date in attached to the actions that are dispatched

src/Service.Host/client/src/components/parts/Part.js
lewisrenfrew: You mean just to reduce the sheer volume of props being passed into these Tab Components? I think I actually prefer really explicit props like this but it does get a bit unwieldly

richardiphillips:
I've only quickly skimmed this. I don't know how it felt writing it but I quite like the look of it. 
Would be nice to get a good way of testing it. Feels similar to testing the redux reducers.

richardiphillips:
> I've only quickly skimmed this. I don't know how it felt writing it but I quite like the look of it.
> Would be nice to get a good way of testing it. Feels similar to testing the redux reducers.

(Just realised that you had already noted the similarity to redux reducers at the top. )

lewisrenfrew:
> I've only quickly skimmed this. I don't know how it felt writing it but I quite like the look of it.
> Would be nice to get a good way of testing it. Feels similar to testing the redux reducers.

I was flip flopping between writing react-testing-library 'test only what the user sees' tests, and reducer function unit tests. In the end I decided to extract the reducer into its own file, making sure it's just a pure function, and then unit test it as normal. 

I thought having all the state management logic that was ported over from the oracle form in one nicely tested function was a good idea.

@atcampbell might disagree though. It's against the way of the dodds. The perfect answer might be write both kinds of tests!

Maybe a discussion for a show n tell

richardiphillips:
> > I've only quickly skimmed this. I don't know how it felt writing it but I quite like the look of it.
> > Would be nice to get a good way of testing it. Feels similar to testing the redux reducers.
> 
> I was flip flopping between writing react-testing-library 'test only what the user sees' tests, and reducer function unit tests. In the end I decided to extract the reducer into its own file, making sure it's just a pure function, and then unit test it as normal.
> 
> I thought having all the state management logic that was ported over from the oracle form in one nicely tested function was a good idea.
> 
> @atcampbell might disagree though. It's against the way of the dodds. The perfect answer might be write both kinds of tests!
> 
> Maybe a discussion for a show n tell

Personally I think there's a place for both. Having the display logic tested is great but having the 'engine' of the page separately unit tested seems to add a good degree of confidence and also good documentation of what it's doing. We should be careful not to put too much logic into the client too. It could be tempting if we've got a powerful reducer function there to let business logic seep into it.

PR #101, opened 2021-02-19
Parcel ut pr attempt 2

iaincrawf:
Unfinished view/edit/create, but working search page. Still need front end tests too.

Carriers is now unused as the carriers for parcels are suppliers with the bool ApprovedCarrier - should I delete the carriers module and classes etc, or does anyone know if it'll be used in another part of stores?


richardiphillips:
> Unfinished view/edit/create, but working search page. Still need front end tests too.
> 
> Carriers is now unused as the carriers for parcels are suppliers with the bool ApprovedCarrier - should I delete the carriers module and classes etc, or does anyone know if it'll be used in another part of stores?

If in doubt - delete

src/Facade/Services/ParcelService.cs
```diff
@@ -45,29 +46,50 @@ protected override void UpdateFromResource(Parcel entity, ParcelResource updateR
...
+                        && x.DateCreated < DateTime.Parse(searchTerms.DateCreatedSearchTerm).AddDays(1)));
+        }
+
+        public IResult<IEnumerable<Parcel>> Search(ParcelSearchRequestResource resource)
+        {
+            return new SuccessResult<IEnumerable<Parcel>>(this.Repository.FilterBy(this.SearchExpression(resource)));
+        }
     }
 
     // public class ImportBookFacadeService : IFacadeService<>
```
richardiphillips, line 95: ?

richardiphillips (approved):
Looks good. 
The searches are certainly complicated!
Facade services could do with tests and if you've still got any more logic to add it should probably go down into the domain.

PR #110, opened 2021-03-01
Export rsn service

atcampbell:
Raising this now to look for feedback on routes / process names etc. and the fields that should be present in the rep 25 creation form


andrewfraser73 (changes requested):
Looks good but can we change the rep-25 to something else

andrewfraser73 (approved):
LGTM

PR #115, opened 2021-03-12
refactor requisition processes

richardiphillips:


lewisrenfrew (approved):
Is the commented out stuff supposed to commented out?

richardiphillips:
> Is the commented out stuff supposed to commented out?

Yeah. I need the infrastructure guys to find out if we can get those label printers onto the print server before I know how we can print the address label so I'm leaving the code to get the relevant data in but commented for now.

PR #123, opened 2021-04-06
Part and dept stock pallets fixes

lewisrenfrew:


src/Service.Host/client/src/components/parts/tabs/GeneralTab.js
```diff
@@ -177,6 +178,7 @@ function GeneralTab({
                     loading={productAnalysisCodesSearchLoading}
                     fetchItems={searchProductAnalysisCodes}
                     links={false}
+                    disabled={!!salesPartNumber}
```
richardiphillips, line 181: is !! the same as  [nothing]

src/Persistence.LinnApps/Repositories/PartRepository.cs
```diff
@@ -37,6 +37,15 @@ public Part FindById(int key)
                 .Include(p => p.MechPartSource)
                 .ThenInclude(m => m.MechPartManufacturerAlts)
                 .ToList().FirstOrDefault();
+            
+            if (result != null)
+            {
+                result.SalesPart =
```
richardiphillips, line 43: must admit i'm not sure this should be in the part repo. Not sure the extra call is warranted for the vast majority of calls that we'll do to parts. On the other hand I get why you've done it.

richardiphillips (approved):
Can take or leave the other comments but I would definitely change all name references to sales part to be sales article. The sales part is the replacement construct in the 'new world' products solution so the naming could cause some confusion. Sales article is the version that is only written back for use by the old world systems (and, excepting a couple of fields that weren't replicated, it isn't ever changed except by the write-back from products)

src/Persistence.LinnApps/Repositories/PartRepository.cs
lewisrenfrew: I think you're right. This GetById is taking long enough to return as it is. I will move this step to the domain service update method

src/Service.Host/client/src/components/parts/tabs/GeneralTab.js
lewisrenfrew: It's a thing I sometimes do to quickly convert any javascript type to a bool. Anything falsy becomes false and anything truthy becomes true. Do you think I should do it more explicitly?

richardiphillips: No that's pretty cool. I'm so used to just using the variable directly as the truthy thing in js that it looked strange.
Does the component proptypes complain about being passed a string? I suppose it (quite reasonably) will.

lewisrenfrew: Exactly... "disabled" expects a bool. The alternative to this is Boolean(value), but !! wins for me since it's less typing

PR #132, opened 2021-04-23
Tpk v1

lewisrenfrew:
not tested in real life yet but just wanted to merge something before I racked up even more conflicts


src/Domain.LinnApps/Tpk/TpkService.cs
```diff
@@ -0,0 +1,125 @@
...
+            var whatToWand = this.whatToWandService.WhatToWand(from.LocationId).ToList();
+            
+            this.tpkPack.UpdateQuantityPrinted(from.FromLocation, out var updateQuantitySuccessful);
+            
+            if (!updateQuantitySuccessful)
+            {
+                throw new TpkException("Failed in update_qty_printed.");
+            }
+            
+            this.storesOoPack.DoTpk(from.LocationId, from.PalletNumber, DateTime.Now, out var tpkSuccessful);
```
richardiphillips, line 96: variable name is wrong now

src/Proxy/StoresPack.cs
```diff
@@ -68,6 +76,79 @@ public ProcessResult UnAllocateRequisition(int reqNumber, int? reqLineNumber, in
             };
         }
 
+        public void DoTpk(int locationId, int palletNumber, DateTime dateTimeStarted, out bool success)
+        {
+            success = false;
+            using (var connection = this.databaseService.GetConnection())
+            {
+                connection.Open();
+                var cmd = new OracleCommand("stores_oo.do_tpk", connection) //  TODO - Move function to stores_pack
```
richardiphillips, line 85: todo is done?

richardiphillips (approved):
looks awesome

src/Proxy/StoresPack.cs
lewisrenfrew: I meant to ask - were you saying we should copy the actual PLSQL procedures from stores_oo into stores_pack? Or just keep them all in the same C# StoresPack proxy?

richardiphillips: stores_pack is just a wrapper for stores_oo. The stores_oo functions should just be called from in there.
It allows us to grab the messages generated in the session and translate booleans to int results etc

PR #136, opened 2021-04-29
Mps fixes

lewisrenfrew:
- fix and refactor various issues I came across when trying to work out the PART_DATA_SHEETs mystery


richardiphillips (approved):
don't know what 90% is actually doing but the code looks good!

PR #138, opened 2021-05-04
Fix parcel routing + pass back actual error messages from db

iaincrawf:


src/Service/Modules/ParcelsModule.cs
```diff
@@ -40,26 +42,41 @@ private object GetParcels()
...
+                    .WithMediaRangeModel("text/html", ApplicationSettings.Get);
 
-            return this.Negotiate.WithModel(result)
-                .WithMediaRangeModel("text/html", ApplicationSettings.Get);
+            }
+            catch (Exception e)
+            {
+                return new ServerFailureResult<Parcel>($"Error when updating - ${(e.InnerException != null ? e.InnerException.Message : e.Message)}");
+            }
         }
```
lewisrenfrew, line 80: I would be tempted to move the try catches to the facade service and return SuccessResults or ServerFailureResult or similiarly acccordingly from there

lewisrenfrew (approved):

src/Service/Modules/ParcelsModule.cs
iaincrawf: Yeah it did feel a bit wrong in the module, but I'd have to overwrite the shared facadeService code which also doesn't feel right. We haven't passed back actual database errors elsewhere have we? Maybe I should just take these out and leave it passing back the "something went horribly, horribly wrong" or whatever error and we can see the actual issue in redux state, was just thinking it could save us time to get a real error on the screen for users to send

richardiphillips: I agree. Should defo be in the facade service. If it's just to catch db errors that 'should' never happen then you could just take them out. The client can show 500s too

iaincrawf: Do you guys think it's worth writing custom facade add/update methods to send back specific errors, or is our general practice just give 500s and we can figure it out? The example that made me want to was the error below, cause we could then look that up and know it's to do with that person not being allowed to check in parcels without having to try recreating the problem ourselves:
ORA-02291: integrity constraint (DEV.PAR1_AGEN_FK) violated - parent key not found ORA-06512

I'm leaning towards just get rid of it and let it be a 500 or normal "something went wrong" error for the user that we can then investigate

richardiphillips: it sounds like something that should be caught in code rather than relying on the database?

iaincrawf: Looked it up again and it's actually just a constraint checking the person is in auth_users not anything parcel specific - think I only got the error in the first place due to using bad sys data and shouldn't happen in live, have removed the try catches

PR #148, opened 2021-05-18
use FacadeFilterService

richardiphillips:
Just a suggestion PR. 

I noticed you extended FacadeService but I think the standard FacadeFilterService does exactly what you want anyway. I haven't tested it so check it out and see if it does what you want...


iaincrawf (approved):
Tried locally and seems all good, thanks!

PR #156, opened 2021-05-25
remove requirement for datasheets at point of creation as per ECITs r…

lewisrenfrew:
…equest


richardiphillips (approved):
looks good (although this was a feature that they specifically added because people were forgetting to add them )

PR #164, opened 2021-06-04
Shipfile optimisations

lewisrenfrew:
- splits the workload across several requests as opposed to waiting for them all to complete in one request. 
- seems to speed it up since requests can be processed in parallel
- I also changed it so that the chrome instance is launched  during Dependency Injection as a singleton as opposed to in the PdfService for every request. I put this in the IoC ServiceModule although I'm not sure thats correct?
- I'm hoping .Net is smart enough not to crash the container with too much work...


iaincrawf (approved):
sending the emails separately seems a good shout, lgtm

PR #166, opened 2021-06-04
more impbook backend

iaincrawf:
Something with ledger periods doesn't build so I need to fix that on Tues before I merge, but thought I'd put the rest up in case anyone gets a chance to pr on Monday


src/Domain.LinnApps/ImportBooks/LedgerPeriod.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Linn.Stores.Domain.LinnApps.ImportBooks
+{
+    public class LedgerPeriod
```
richardiphillips, line 3: i'd move this out of the importbooks folder. definitely a general concern.

src/Facade/Services/ImportBookTransactionCodeService.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Linn.Stores.Facade.Services
+{
+    using System.Collections.Generic;
+
+    using Linn.Common.Facade;
+    using Linn.Common.Persistence;
+    using Linn.Stores.Domain.LinnApps.ImportBooks;
+
+    public class ImportBookTransactionCodeService : IImportBookTransactionCodeService
```
richardiphillips, line 9: why not just use the standard facade service? then you don't need to write any of this code and we have more functionality available if we need it for free.

src/Facade/Services/ImportBookTransportCodeService.cs
```diff
@@ -0,0 +1,25 @@
...
+    using Linn.Stores.Domain.LinnApps.ImportBooks;
+
+    public class ImportBookTransportCodeService : IImportBookTransportCodeService
+    {
+        private readonly IRepository<ImportBookTransportCode, int> transportCodeRepository;
+
+        public ImportBookTransportCodeService(IRepository<ImportBookTransportCode, int> transportCodeRepository)
+        {
+            this.transportCodeRepository = transportCodeRepository;
+        }
```
richardiphillips, line 16: use the standard facade service?

tests/Unit/Facade.Tests/ImportBookFacadeServiceTests/WhenCreating.cs
```diff
@@ -0,0 +1,337 @@
...
+        private ImportBookResource resource;
+
+        private ImportBook from;
+
+        private IResult<ImportBook> result;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.resource = new ImportBookResource()
```
richardiphillips, line 33: don't need ()

tests/Unit/Facade.Tests/ImportBookFacadeServiceTests/WhenCreating.cs
```diff
@@ -0,0 +1,337 @@
...
+                                                                            EntryDate = this.now.AddDays(-6)
+                                                                                .ToString("o"),
+                                                                            Reference = "hocus pocus",
+                                                                            Duty = 33,
+                                                                            Vat = 44
+                                                                        }
+                                                                }
+                                };
+
+            this.ImportBookRepository.FindById(Arg.Any<int>()).Returns(this.from);
```
richardiphillips, line 181: don't think you need the repo mock

richardiphillips (approved):
:thumbsup:

src/Facade/Services/ImportBookTransactionCodeService.cs
iaincrawf: Yeah should've used it, done now. I think it was because I did the exchange rate one first where I was calling the domain and just didn't think the next ones through properly

src/Domain.LinnApps/ImportBooks/LedgerPeriod.cs
iaincrawf: Cool. On ledgerPeriod - I can't get it to run, it's complaining that it can't map the table into the repository. When I do desc ledger_periods, period_number is mapped as just a NUMBER, with no limit so guess that will be the max oracle size? I wondered after int was failing, so tried long and decimal but still not working. Do you know of anything we normally have used or is it likely I've just missed something else?

richardiphillips: Can't see anything obviously wrong with that. If you're still stuck i'll pull it down and have a look

iaincrawf: If you get a chance that'd be much appreciated, it seems like such a simple table and I can't see anything silly that's wrong with it so far. Wondered if it was that the sys/dev db was different but it's not that either

src/Facade/Services/ImportBookTransactionCodeService.cs
richardiphillips: i remember lewis mapped it as a view in the tax stuff. you could try that for now as we'll never update that table

src/Domain.LinnApps/ImportBooks/LedgerPeriod.cs
richardiphillips: i remember lewis mapped it as a view in the tax stuff. you could try that for now as we'll never update that table

iaincrawf: Was all a wild goose chase, sorry! I kept missing that I was trying to register LedgerPeriod to the IRepository, not LedgerPeriodRepository to it. Big time Doh. Seems happy as a table/dbset now though even with just an int

PR #179, opened 2021-06-22
Consignment pallets and items

richardiphillips:


andrewfraser73 (approved):
LGTM

However there is talk about maybe having double the customs entries on consignments (or on something else) but we can do that if they ever decide what way they want to go - either way its better to have it here where we can edit it in the real world

PR #186, opened 2021-07-12
fix for entering export customs entries

andrewfraser73:
small bug fixes to make export returns work for entering export customs entries after they have generated inter company invoices


iaincrawf (approved):
lovely juvly

PR #188, opened 2021-07-15
impbook IPR report v0

iaincrawf:
Can't figure out how to get the impbook Id into the href to link to the import book ut, because I need the rowId to be a combo of it and orderDetails.lineNumber so that it's unique and shows all rows.

Also currently can't get impbook ut on apps so figure I missed something routing wise but can't see what, have made a change to the impbook module which may or may not help


src/Domain.LinnApps/Reports/ImportBookReportService.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public ResultsModel GetIPRReport(DateTime from, DateTime to)
+        {
+            var iprImpbooks = this.impbookRepository.FilterBy(
+                x =>
+
+                    // x.IprCpcNumber.HasValue && x.IprCpcNumber.Value == 13 &&
+                    // todo check if I can just use that ^, not sure it'll be populated reliably 
+                    x.OrderDetails.Any(z => z.CpcNumber.HasValue && z.CpcNumber.Value == this.IprCpcNumberId)
```
andrewfraser73, line 33: as discussed before 13 might work but not necessarily be future proof.  it gets us going. The most probably mistake that will happen is Rhona or someone will pick the CPC number that is close but wrong (I can see 2 this year using the extra 0 version) however all my processing will also only really work with 13

src/Facade/Services/ImportBookReportReportFacadeService.cs
```diff
@@ -0,0 +1,40 @@
...
+            DateTime from;
+            DateTime to;
+            try
+            {
+                from = DateTime.Parse(resource.FromDate);
+                to = DateTime.Parse(resource.ToDate);
+            }
+            catch (Exception)
+            {
+                return new BadRequestResult<ResultsModel>("Invalid dates supplied to assembly fails details report");
```
andrewfraser73, line 34: think this must be a copy paste from assembly fails as should prob say impbook ipr report

src/Service.Host/client/src/components/reports/ImpbookIprReport.js
```diff
@@ -0,0 +1,72 @@
...
+
+const handleBackClick = (history, options) => {
+    const uri = `/logistics/import-books/ipr?fromDate=${encodeURIComponent(
+        options.fromDate
+    )}&toDate=${encodeURIComponent(options.toDate)}`;
+
+    history.push(uri);
+};
+
+const AssemblyFailsDetails = ({ reportData, loading, history, options, config }) => (
```
andrewfraser73, line 20: lots of references to assemblyfails details might want to find/replace them to impbookipr

src/Service.Host/client/src/components/reports/ImpbookIprReport.js
```diff
@@ -0,0 +1,72 @@
...
+                )}
+            </Grid>
+            <Grid item xs={12}>
+                <BackButton backClick={() => handleBackClick(history, options)} />
+            </Grid>
+        </Grid>
+    </Page>
+);
+
+AssemblyFailsDetails.propTypes = {
```
andrewfraser73, line 55: same as before

andrewfraser73 (approved):
might want to change a couple of the assemblyfail refs

rest looks alright but might double check the report results when its live on production

src/Domain.LinnApps/Reports/ImportBookReportService.cs
iaincrawf: Yeah, we could pull all of the cpc numbers in easily enough but even then, we either use the right one with id 13, or include the other similar ones and maybe get dodgy results! Maybe I can put an 'are you sure this is the right cpc number/ is this IPR?' popup in the new form to hopefully stop the wrong one being used

src/Facade/Services/ImportBookReportReportFacadeService.cs
iaincrawf: Yeah sorry, lots of copy and pasting for speed - should have checked through it a bit better though!

src/Domain.LinnApps/Reports/ImportBookReportService.cs
andrewfraser73: i would just leave it is until we get the report working.  it is a future worry that doesn't stop you doing what you need to do

PR #189, opened 2021-07-16
More fields on IPR report and couple of dropdowns properly populated on impbook

iaincrawf:
IPR report is still missing tests and the entire export function, and impbook is still very much incomplete. This pr will get through the rest of the fields that Yvonne wants on the IPR report for now though

### DIFF
--- a/src/Persistence.LinnApps/Repositories/ImportBookRepository.cs
+++ b/src/Persistence.LinnApps/Repositories/ImportBookRepository.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Linq;
     using System.Linq.Expressions;
+    using System.Security.Cryptography.X509Certificates;
 
     using Linn.Common.Persistence;
     using Linn.Stores.Domain.LinnApps.ImportBooks;
@@ -49,7 +50,9 @@
 
         public IQueryable<ImportBook> FilterBy(Expression<Func<ImportBook, bool>> expression)
         {
-            return this.serviceDbContext.ImportBooks.Where(expression).Include(y => y.OrderDetails);
+            return this.serviceDbContext.ImportBooks.Where(expression)
+                .Include(y => y.OrderDetails)
+                .Include(x => x.InvoiceDetails);
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/Persistence.LinnApps/Repositories/ImportBookRepository.cs
```diff
@@ -3,6 +3,7 @@
     using System;
     using System.Linq;
     using System.Linq.Expressions;
+    using System.Security.Cryptography.X509Certificates;
```
richardiphillips, line 6: probably not needed

src/Service.Host/client/src/components/importBooks/tabs/ImpBookTab.js
```diff
@@ -417,9 +417,11 @@ function ImpBookTab({
                 </Grid>
 
                 <Grid item xs={6}>
-                    {/* todo implement delivery term code stuff  */}
                     <Dropdown
-                        items={[]}
+                        items={deliveryTerms.map(e => ({
```
richardiphillips, line 421: might be nice to put these items in a variable above

richardiphillips (approved):

src/Service.Host/client/src/components/importBooks/tabs/ImpBookTab.js
iaincrawf: Do you mean map them somewhere else, like in the container when they're selected? Or to store them in something like a localDeliveryTerms variable? I'd prefer not to store them locally as I've got a few like this and adding consts & useEffects feels a bit cluttered when it's not adding anything

richardiphillips: i never thought of doing it in the container. That actually might be quite nice if the items aren't used for anything else. I just meant in a const as i think it makes the dropdown declaration a bit cleaner.

lewisrenfrew: +1 for mapping arrays passed as props into the form actually used by the component in the container

iaincrawf: Yeah I've got a few where the items aren't used for anything except dropdowns, I'll move the mappings to the container and then can just have nice "items={deliveryTerms}" in the components

PR #190, opened 2021-07-19
handle exception properly so stores guys can see the Cannot Delete St…

lewisrenfrew:
…ock Locators messages

### DIFF
--- a/src/Facade/Services/StockLocatorsFacadeService.cs
+++ b/src/Facade/Services/StockLocatorsFacadeService.cs
@@ -6,6 +6,7 @@
     using Linn.Common.Facade;
     using Linn.Common.Persistence;
     using Linn.Common.Proxy.LinnApps;
+    using Linn.Stores.Domain.LinnApps.Exceptions;
     using Linn.Stores.Domain.LinnApps.StockLocators;
     using Linn.Stores.Resources.RequestResources;
     using Linn.Stores.Resources.StockLocators;
@@ -35,7 +36,14 @@
         public IResult<StockLocator> Delete(StockLocatorResource resource)
         {
             var toDelete = this.repository.FindById(resource.Id);
-            this.domainService.DeleteStockLocator(toDelete, resource.UserPrivileges);
+            try
+            {
+                this.domainService.DeleteStockLocator(toDelete, resource.UserPrivileges);
+            }
+            catch (StockLocatorException ex)
+            {
+                return new BadRequestResult<StockLocator>(ex.Message);
+            }
             return new SuccessResult<StockLocator>(toDelete);
         }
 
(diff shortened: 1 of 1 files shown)


src/Facade/Services/StockLocatorsFacadeService.cs
```diff
@@ -35,7 +36,14 @@ public class StockLocatorsFacadeService :
...
-            this.domainService.DeleteStockLocator(toDelete, resource.UserPrivileges);
+            try
+            {
+                this.domainService.DeleteStockLocator(toDelete, resource.UserPrivileges);
+            }
+            catch (StockLocatorException ex)
+            {
+                return new BadRequestResult<StockLocator>(ex.Message);
+            }
             return new SuccessResult<StockLocator>(toDelete);
```
richardiphillips, line 47: add blank line?

richardiphillips (approved):
lgtm

lewisrenfrew:
> lgtm

yeah... It's better than them just getting a 500. Still can't see how or why dept stock stock locators are ending up with dependent req moves existing. 

It seems like an odd coincidence that this justt seems to have started happening since I did the new dept-stock-ut, but the errors occur on the old oracle form too and I definitely didn't change anything there so I'm not convinced it's something I got wrong.

Every time the stores guys see this error we just end up deleting any dependent req_move anyway. I could just add a check in before I do the delete stock locator here to preemptively delete any dependent req_moves.. Do you think that's a good idea?

richardiphillips:
> > lgtm
> 
> yeah... It's better than them just getting a 500. Still can't see how or why dept stock stock locators are ending up with dependent req moves existing.
> 
> It seems like an odd coincidence that this justt seems to have started happening since I did the new dept-stock-ut, but the errors occur on the old oracle form too and I definitely didn't change anything there so I'm not convinced it's something I got wrong.
> 
> Every time the stores guys see this error we just end up deleting any dependent req_move anyway. I could just add a check in before I do the delete stock locator here to preemptively delete any dependent req_moves.. Do you think that's a good idea?

Is it happening a lot?

lewisrenfrew:
> > > lgtm
> > 
> > 
> > yeah... It's better than them just getting a 500. Still can't see how or why dept stock stock locators are ending up with dependent req moves existing.
> > It seems like an odd coincidence that this justt seems to have started happening since I did the new dept-stock-ut, but the errors occur on the old oracle form too and I definitely didn't change anything there so I'm not convinced it's something I got wrong.
> > Every time the stores guys see this error we just end up deleting any dependent req_move anyway. I could just add a check in before I do the delete stock locator here to preemptively delete any dependent req_moves.. Do you think that's a good idea?
> 
> Is it happening a lot?

Not enough for me to notice any patterns yet anyway. Maybe 4 times in the last few months but I'm not sure how often they actually delete stock locators using this form. 
I've got Mark and Colin on watch for it though and they'll get in touch with me every time it happens. I'll just leave it for now and see if anything develops

PR #194, opened 2021-07-21
IPR impbooks - Export function & tests

iaincrawf:


src/Facade/Services/ImportBookReportFacadeService.cs
```diff
@@ -0,0 +1,57 @@
...
+            this.reportService = reportService;
+        }
+
+        public IResult<ResultsModel> GetImpbookIPRReport(IPRSearchResource resource)
+        {
+            DateTime from;
+            DateTime to;
+            try
+            {
+                from = DateTime.Parse(resource.FromDate);
```
lewisrenfrew, line 27: could do DateTime.TryParse() here instead of a try/catch. Maybe a tad cleaner but I don't have any strong feelings. @richardiphillips likes a TryParse() so he might

src/Facade/Services/ImportBookReportFacadeService.cs
```diff
@@ -0,0 +1,57 @@
...
+                return new BadRequestResult<ResultsModel>("Invalid dates supplied to impbook IPR report");
+            }
+
+            return new SuccessResult<ResultsModel>(this.reportService.GetIPRReport(from, to));
+        }
+
+        public IResult<IEnumerable<IEnumerable<string>>> GetImpbookIprReportExport(IPRSearchResource resource)
+        {
+            DateTime from;
+            DateTime to;
```
lewisrenfrew, line 41: if you do go down the TryParse route you could ditch these declarations and just do

`DateTime.TryParse(dateString, out var from)`

and same for to

lewisrenfrew (approved):

src/Facade/Services/ImportBookReportFacadeService.cs
iaincrawf: I really like how clean this is, but just writing a test to see what it does and it seems to pass in "01/01/0001" when fed nonsense. Are we ok with that cause it's basically not going to get called without the right values? I guess worst case is just that it could do a slow select which doesn't really matter

richardiphillips: catch the bool it returns and you'll know if it was valid or not

PR #195, opened 2021-07-22
Goods in ut 3

lewisrenfrew:


src/Domain.LinnApps/GoodsIn/GoodsInService.cs
```diff
@@ -136,5 +197,23 @@ public class GoodsInService : IGoodsInService
 
             return result;
         }
+
+        public ProcessResult ValidatePurchaseOrderQty(
+            int orderNumber, 
+            int qty,
+            int? orderLine)
+        {
+            var valid = this.storesPack.ValidOrderQty(orderNumber, orderLine ?? 1, qty, out _, out _);
```
iaincrawf, line 206: What does out _ do?

src/Service.Host/client/src/components/goodsIn/GoodsInUtility.js
```diff
@@ -94,6 +114,15 @@ function GoodsInUtility({
                         value={formData.qty}
                         label="Qty"
                         propertyName="qty"
+                        textFieldProps={{
+                            onBlur: () =>
+                                validatePurchaseOrderBookInQty(
+                                    `orderNumber=${formData.purchaseOrderNumber}&qty=${
+                                        formData.qty
+                                    }&orderLine=${1}&`,
```
iaincrawf, line 122: not sure if more params added elsewhere or if a wee extra '&'

iaincrawf (approved):
Looks good to me

src/Domain.LinnApps/GoodsIn/GoodsInService.cs
lewisrenfrew: Nothing really. I think its just a way of denoting that a variable will be unused. I would guess under the hood it stops the runtime allocating memory for that variable

PR #196, opened 2021-07-22
EU Impbooks report v1, IPR report improvements

iaincrawf:
Need to do more domain tests & also need to work out what fields connect the intercompany invoices to impbooks & bring them in for two or three report columns, but think it'll be more useful to just get them out as blank fields for now so that Yvonne has the full ipr report with columns in the right order to check while she's wfh today

This also has switches to toggle between ipr and non ipr impbooks, and eu and rest of world impbooks


src/Domain.LinnApps/ImportBooks/ImportBook.cs
```diff
@@ -122,5 +122,8 @@ public class ImportBook
         public IList<ImportBookOrderDetail> OrderDetails { get; set; }
 
         public IList<ImportBookPostEntry> PostEntries { get; set; }
+
+        public Supplier FullSupplier { get; set; }
+        public Supplier FullCarrier { get; set; }
```
richardiphillips, line 127: add blank line? What does fullsupplier or fullcarrier mean out of interest

richardiphillips (approved):
looking good

src/Domain.LinnApps/ImportBooks/ImportBook.cs
iaincrawf: Just that it's the full supplier object which isn't normally on an impbook, it only stores supplierId. Maybe I should just call them Supplier and Carrier? For some reason I wanted to differentiate, I also only include/pull them in for filterBy and not findById

richardiphillips: yeah i'd call them supplier and carrier.

PR #198, opened 2021-07-27
More dept stock ut improvements

lewisrenfrew:
- uses result of the previous request cached in redux store (if present) as opposed to refetching the data every time. Makes using the utility a lot nicer as the slow load time only happens on initial load and then clicking back and searching for a new part is instantaneous
- temporary hacky fix for the dept stock pallet issue so Mark and co can get on with their lives. I thought if I just break the link between the stock locator and its dependent req move as opposed to deleting it entirely (and also log the event in the req_move's remarks field) then I'll at least be able to watch for when the issue occurs and come back to it later


iaincrawf (approved):
Sounds good to me

PR #199, opened 2021-07-27
IPR fields from intercompany inv

iaincrawf:
Intercompany inv ref and a customs value set from intercompany invoice and export return details.

Eventually worked out that I had to set a Principal key to get intercompany inv by rsn which isn't its primary key, seems to work but haven't done that before.

Should maybe spend some more time covering more of the domain service with tests? Currently have a good bit of the IPR report covered though which is ok I guess


tests/Unit/Domain.LinnApps.Tests/Domain.LinnApps.Tests.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>netcoreapp2.2</TargetFramework>
     <AssemblyName>Linn.Stores.Domain.LinnApps.Tests</AssemblyName>
     <RootNamespace>Linn.Stores.Domain.LinnApps.Tests</RootNamespace>
+    <LangVersion>latest</LangVersion>
```
iaincrawf, line 7: Should I include this? I Had to set it to un-break some tests using named params

src/Domain.LinnApps/Reports/ImportBookReportService.cs
```diff
@@ -328,186 +354,197 @@ private void AddExportColumns(SimpleGridLayout reportLayout)
...
-        ICollection<CalculationValueModel> values,
-        ImportBook impbook,
-        ImportBookOrderDetail orderDetail,
-        bool isInEU = false)
+            ICollection<CalculationValueModel> values,
+            ImportBook impbook,
+            ImportBookOrderDetail orderDetail,
+            bool isInEU)
         {
+            var exportReturnDetail = (ExportReturnDetail)null;
```
iaincrawf, line 369: I don't think I like this, not sure it's the best. Anyone think it'd be better just initialising it normally and then checking the same conditions as the if below, instead of checking for exportReturnDetail being null later on? Or any other suggestions if anyone is bothered

src/Persistence.LinnApps/Repositories/InterCompanyInvoiceRepository.cs
```diff
@@ -30,5 +30,20 @@ public IQueryable<InterCompanyInvoice> FindAll()
...
+
+        public void Add(InterCompanyInvoice entity)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Remove(InterCompanyInvoice entity)
+        {
+            throw new NotImplementedException();
+        }
```
lewisrenfrew, line 47: Why change to an IRepository if none of the new methods are implemented? Or is this for the future?

tests/Unit/Domain.LinnApps.Tests/Domain.LinnApps.Tests.csproj
lewisrenfrew: no idea

lewisrenfrew (approved):

src/Persistence.LinnApps/Repositories/InterCompanyInvoiceRepository.cs
iaincrawf: Can I still pull them in with exportReturnDetails as an IQueryRepo? If so then you make a v good point and I'll change back, I did it while I was trying everything under the sun to try and get that relationship working and forgot to check if it was actually needed

iaincrawf: Actually will check with Andrew whether he's going to be doing anything with intercompany invoices in stores that will use it soon, otherwise will get revert (unless it breaks but think you're right should be fine)

src/Domain.LinnApps/Reports/ImportBookReportService.cs
lewisrenfrew: I didn't see this comment since github hides larges diffs.. 
Yeah - this seems like an odd thing to do. If you need to initialise exportReturnDetail with a null value it doesn't make much sense to make it an implicitly typed var

iaincrawf: I don't know what I was thinking with this yesterday, I'll just change var to ExportDetailReturn and set it to null in my next impbook pr

PR #200, opened 2021-07-28
close consignment

richardiphillips:


src/Messaging/Dispatchers/PrintConsignmentNoteDispatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Serialization;
+
+    public class PrintConsignmentNoteDispatcher : IPrintConsignmentNoteDispatcher
+    {
+        private const string ContentType = "application/json";
+
+        private readonly string routingKey = "orawin.consignment-note.print";
+
```
lewisrenfrew, line 17: any readon why one is a const and one is a readonly?

lewisrenfrew (approved):

src/Messaging/Dispatchers/PrintConsignmentNoteDispatcher.cs
richardiphillips: not really

andrewfraser73 (approved):
Export book pack on close integration looks good.

PR #212, opened 2021-08-19
Impbook - most of front end

iaincrawf:
Still need to 
- (_done_) implement privileges check properly once I find out who should be able to edit
- write a big check to make sure all of the required fields are filled out before allowing save
- (_done_) implement the exchange rates bit, maybe just a search lookup thing like the others
- implement the invoice details function, maybe as a popup dialog box
- write some tests
- Make/edit shared component for the searchInput + clear button combo I've used here and on parcel ut, maybe make it look neater too

Going to send this to Rhona to get feedback etc once it's through, maybe ask about the invoice details too


src/Service.Host/client/src/containers/importBooks/tabs/ImpBookTab.js
```diff
@@ -46,21 +48,40 @@ const mapStateToProps = state => ({
...
+        !props.transactionCodes ||
+        props.transactionCodes.length === 0 ||
+        !props.deliveryTerms ||
+        props.deliveryTerms.length === 0 ||
+        !props.ports ||
+        props.ports.length === 0 ||
+        !props.countries ||
+        props.countries.length === 0
+    ) {
+        dispatch(employeesActions.fetch());
```
lewisrenfrew, line 77: I would maybe do each of these ifs specific to each dispatch,
or if that's too many lines of code do
```
(!props.employees || props.employees.length === 0 ) && dispatch(employeesActions.fetch());
.
.
etc
```

lewisrenfrew (approved):
Looks good to me. I'd address the todos or remove them and stick them in an Issue or something to keep the production code clean before merging

src/Service.Host/client/src/containers/importBooks/tabs/ImpBookTab.js
iaincrawf: I was kinda thinking if any of them aren't there, then the data isn't from this page, so could be wrong and they all should be loaded again. I do quite like the && syntax wise too though, might do that

PR #214, opened 2021-08-20
First Impbook alterations after Rhona chat

iaincrawf:
- Removed fields that Rhona said she didn't need
- Added table to first page for invoice details
- Added in code to do calculation on totals (ie check that invoice details total == total from order details) to make sure that everything is balanced and comes out as 0, displayed in fields at top of order details

Still to do auto printing, reprinting function and autopopulation of parcel ut fields, autopopulation of order details fields from RSNs etc, plus work out where post duty function posts things to and write that too. + tests & maybe couple of shared component additions/edits

### DIFF
--- a/src/Service.Host/client/src/components/importBooks/ImportBook.js
+++ b/src/Service.Host/client/src/components/importBooks/ImportBook.js
@@ -180,11 +180,31 @@ function ImportBook({
         });
     };
 
+    const handleUpdateInvoiceDetails = details => {
+        setEditStatus('edit');
+        dispatch({
+            type: 'invoiceDetailsUpdate',
+            details
+        });
+    };
+
     const allowedToEdit = () => {
         if (!(privileges.length < 1)) {
             return privileges.some(priv => priv === 'import-books.admin');
         }
         return false;
+    };
+
+    const totalInvoiceValue = () => {
+        let total = `${state.impbook.importBookInvoiceDetails?.reduce(
+            (a, v) => a + v.invoiceValue,
+            0
+        )}`;
+        if (total) {
+            total = `${parseFloat(total).toFixed(2)}`;
+        }
+
+        return total;
     };
 
     return (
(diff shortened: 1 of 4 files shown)


src/Service.Host/client/src/components/importBooks/ImportBook.js
```diff
@@ -180,13 +180,33 @@ function ImportBook({
...
         return false;
     };
 
+    const totalInvoiceValue = () => {
+        let total = `${state.impbook.importBookInvoiceDetails?.reduce(
+            (a, v) => a + v.invoiceValue,
+            0
+        )}`;
+        if (total) {
+            total = `${parseFloat(total).toFixed(2)}`;
```
lewisrenfrew, line 204: what's parseFloat doing?

src/Service.Host/client/src/components/importBooks/tabs/OrderDetailsTab.js
```diff
@@ -61,49 +61,74 @@ function OrderDetailsTab({
...
+        const orderDetailsWeightTotal = orderDetails?.reduce((a, v) => a + v.weight, 0);
+
+        let remaining = `${weight - orderDetailsWeightTotal}`;
+
+        if (remaining) {
+            remaining = `${parseFloat(remaining).toFixed(2)}`;
+        }
+
+        return remaining || 0;
+    };
```
lewisrenfrew, line 98: I could be wrong, but you could risk running into floating point precision errors here doing totals etc. in Javascript. I tend to either do sums in the back end with c# decimal objects, or us a number javascript library that'll take care of this stuff for you. decimal.js is already in the project if you do decide to go down the js route

lewisrenfrew (approved):

src/Service.Host/client/src/components/importBooks/tabs/OrderDetailsTab.js
iaincrawf: I think you're probably right with that, it seemed ok at a glance but I know js does some unpredictable things with numbers. I'm going to put this through as I've directed some folk to this screen as an option to view impbooks, but I'll definitely change it and try using decimal.js in my next pr. Don't think there's any immediate danger here as only Yvonne and Rhona have edit permission and both know it isn't complete yet, so nothing will be being saved

PR #218, opened 2021-08-31
Move items

richardiphillips:
Palletise cartons and items.
Cartonise items. 
Recalculate functions. 

Manually type consignment number in wand for shug.


src/Service.Host/client/src/components/consignments/ItemsTab.js
```diff
@@ -87,6 +101,206 @@ function ItemsTab({
...
+        let weight = 0;
+        const setValues = item => {
+            if (item.itemType === 'C'){
+                weight += item.itemBaseWeight;
+            } else {
+                weight += item.weight;
+            }
+        };
+
+        // eslint-disable-next-line prettier/prettier
```
iaincrawf, line 248: what doesn't it like? I've never actually used the disable-next-line thing before

iaincrawf (approved):
Looks good from my limited consignment-y knowledge. Also something about manually checking key-press codes that I love for some reason

src/Service.Host/config-example.env
```diff
@@ -6,11 +6,11 @@ RABBIT_SERVER=localhost
...
-APP_ROOT=http://localhost:61798
+APP_ROOT=https://localhost:61798
 AUTHORITY_URI=https://www.linn.co.uk/auth/
-PROXY_ROOT=http://app.linn.co.uk
+PROXY_ROOT=https://app.linn.co.uk
 SMTP_HOSTNAME=smarthost
 SHIPFILES_FROM_ADDRESS=someone@something.co.uk
 SHIPFILE_TEMPLATE_PATH=./views/ShipfilePdfTemplate.html
-HTML_TO_PDF_API_CONVERSION_ENDPOINT=http://localhost:5000/convert/html
+HTML_TO_PDF_API_CONVERSION_ENDPOINT=https://localhost:5000/convert/html
```
lewisrenfrew, line 15: for when you update the real configs in S3
For some reason the sys healtcheck for the pdf service (on http and https) is bad gatewaying, but the real one works fine on https, e.g.
https://app.linn.co.uk/pdf-service/ping
so this shouldn't break anything

richardiphillips: i've already switched stores over in real life. Now if anyone links out to an non https friendly solution they'll have to go and fix it first!

PR #220, opened 2021-09-02
Kardex labels

lewisrenfrew:


src/Service.Host/client/src/components/goodsIn/GoodsInUtility.js
```diff
@@ -123,15 +126,18 @@ function GoodsInUtility({
 
     useEffect(() => {
         if (bookInResult?.message) {
-            setMessage({ error: false, text: bookInResult.message, success: bookInResult.success });
+            setMessage({ error: !bookInResult?.success, text: bookInResult.message });
```
iaincrawf, line 129: What does the '!' do here? Also do we need the '?' before success since you've checked bookInResult at the if above?

iaincrawf (approved):

src/Service.Host/client/src/components/goodsIn/GoodsInUtility.js
lewisrenfrew: it's an error if not success, hence the !. And yeah I think that ? is redundant

iaincrawf: Ahhh yeah of course, for some reason I wasn't reading it as a "not". D'oh!

PR #225, opened 2021-09-14
fix data grid version until we move to the release version

richardiphillips:


src/Service.Host/.babelrc
```diff
@@ -13,8 +13,8 @@
     [
       "@babel/plugin-proposal-class-properties",
       {
-        "loose": true
+        "loose": false
       }
```
lewisrenfrew, line 17: you can just omit this entirely since false is the default . Doesn't really matter though

lewisrenfrew (approved):

PR #228, opened 2021-09-15
update data grid

richardiphillips:


src/Service.Host/client/src/components/StockMove.js
```diff
@@ -177,16 +177,16 @@ function StockMove({
         }
     };
 
-    const handleSelectRow = row => {
-        setSelectedRow(availableStock[row.rowIds[0]]);
+    const handleSelectRow = rows => {
+        setSelectedRow(availableStock[rows[0]]);
```
iaincrawf, line 181: So does it now just give you a list of rowIds on select?

iaincrawf (approved):
Not overly familiar with datagrids to begin with bit looks good if they still work! If it does return a list of rowIds instead of a list of row objects could maybe rename rows to rowIds but lgtm

src/Service.Host/client/src/components/StockMove.js
richardiphillips: yep and doesn't stringify them anymore

PR #229, opened 2021-09-16
Packing list

richardiphillips:


src/Domain.LinnApps/Consignments/ConsignmentService.cs
```diff
@@ -103,6 +107,100 @@ public ProcessResult PrintConsignmentDocuments(int consignmentId, int userNumber
...
+        }
+
+        private bool CanBePutOnPallet(ConsignmentItem consignmentItem)
+        {
+            if (consignmentItem.PalletNumber.HasValue)
+            {
+                return false;
+            }
+
+            return (consignmentItem.ItemType == "I" && consignmentItem.ContainerNumber is null)
```
iaincrawf, line 200: Hadn't realised we could write "is null" in c#, nice

iaincrawf (approved):
Lgtm, do you think it's worth any front end testing on any of this stuff, like is there value in even 'renders some fields'? Saw the main consignment reducer is already tested

richardiphillips:
> Lgtm, do you think it's worth any front end testing on any of this stuff, like is there value in even 'renders some fields'? Saw the main consignment reducer is already tested

yep. i'll try and add some light touch testing at least

PR #230, opened 2021-09-17
consignment print css

iaincrawf:
Not perfect, but seems mostly better?

- got rid of some blank space at top - print style: padding-top-when-not-printing, which has negative margin on print instead of constant margin 
https://github.com/linn/react-components-library/blob/master/src/styles/printStyles.css#L35

- add pageContainer print style to a div containing the consignment page, on print sets the height and width to exactly a4 size
- move style to hide on print onto grids from inner components - seems to get rid of some more empty space at top
Files changed tab looks like carnage but that's all I've done ^

image

image


richardiphillips (approved):
Cool. That certainly looks better at least. 
Thanks

PR #232, opened 2021-09-22
Some bug fixes etc

lewisrenfrew:
was gonna bundle this in with a tpk fix but that's taken a lot longer than anticipated so just PR'ing this tiny parts change for now


src/Service.Host/client/src/components/__tests__/parts/partReducer.test.js
```diff
@@ -73,7 +73,13 @@ describe('partReducer tests', () => {
 
     it('should handle simple field changes', () => {
         expect(
-            partReducer({}, { type: 'fieldChange', fieldName: 'partNumber', payload: 'PART' })
+            partReducer({}, { type: 'fieldChange', fieldName: 'partNumber', payload: ' PART' })
```
richardiphillips, line 76: should this just be 'PART' to demonstrate correct state

src/Service.Host/client/src/components/__tests__/parts/partReducer.test.js
```diff
@@ -73,7 +73,13 @@ describe('partReducer tests', () => {
...
+        ).toEqual({ part: { partNumber: 'PART' } });
+    });
+
+    it('should trim leading/trailing spaces from partNumber', () => {
+        expect(
+            partReducer({}, { type: 'fieldChange', fieldName: 'partNumber', payload: ' PART' })
```
richardiphillips, line 82: maybe test ' PART ' just in case even though it doesn't make any difference?

richardiphillips (approved):
do you think the server side should check this too just in case?

src/Service.Host/client/src/components/__tests__/parts/partReducer.test.js
lewisrenfrew: ah yeah, I must've changed this one by accident and forgot to undo!

src/Service.Host/client/src/components/__tests__/parts/partReducer.test.js
lewisrenfrew: > do you think the server side should check this too just in case?

good call

PR #237, opened 2021-09-23
Consignment component tests

richardiphillips:


src/Service.Host/client/src/components/__tests__/consignments/consignment.test.js
```diff
@@ -0,0 +1,305 @@
+/**
+ * @jest-environment jest-environment-jsdom-sixteen
+ */
```
lewisrenfrew, line 3: I think you only need this weird jsdom-sixteen declaration if you're using async-y stuff

lewisrenfrew (approved):

PR #238, opened 2021-09-24
Impbook

iaincrawf:
Create should be good to go
Oracle impbook print out recreated with new component and print dialog comes up automatically on create (printed automatically on create in oracle)
Component and reducer tests
Unused fields removed from backend

Might still need to autopopulate rsns and some other things in order details (will see if Rhona complains), and need to do that 'post duty' function


src/Service.Host/client/src/components/__tests__/impbooks/Impbook.test.js
```diff
@@ -0,0 +1,281 @@
...
+                privileges={privileges}
+                loading={false}
+                item={item}
+            />
+        );
+    });
+
+    test('invoice values are shown', () => {
+        expect(screen.getByText('99.01')).toBeInTheDocument();
+        expect(screen.getByText('1400')).toBeInTheDocument();
```
lewisrenfrew, line 219: you could do 
<code>
expect(screen.GetByLabelText('blah').value).toBe('99.01')
</code>

if that works. Just cos that tests the correct field has the correct value, as opposed to just checking for the presence of the correct value. I guess in theory your current test might not pick up cases where fields have their values mixed up, if that makes sense

src/Service.Host/client/src/components/importBooks/ImpBookPrintOut.js
```diff
@@ -0,0 +1,386 @@
...
+                </Grid>
+                <Grid item xs={3}>
+                    Weight: <b>{weight}</b>
+                </Grid>
+
+                <Grid item xs={9}>
+                    Transaction Code: <b>{transactionCode}</b>
+                </Grid>
+                <Grid item xs={3}>
+                    Customs Entry Code:{' '}
```
lewisrenfrew, line 163: linter might have introduced {' '}

src/Service.Host/client/src/components/importBooks/ImportBook.js
```diff
@@ -188,6 +163,14 @@ function ImportBook({
...
+        setEditStatus('edit');
+        dispatch({
+            type: 'parcelChange',
+            parcel
+        });
+    };
+
     const allowedToEdit = () => {
         if (!(privileges.length < 1)) {
             return privileges.some(priv => priv === 'import-books.admin');
```
lewisrenfrew, line 176: .includes() would do this in less code

src/Service.Host/client/src/components/importBooks/tabs/ImpBookTab.js
```diff
@@ -188,18 +133,42 @@ function ImpBookTab({
                         onChange={handleFieldChange}
                         type="number"
                         disabled={!allowedToEdit}
+                        data-testid="createdBy"
```
lewisrenfrew, line 136: out of interst: did getByLabelText not work for this Dropdown?

src/Service.Host/client/src/components/importBooks/tabs/ImpBookTab.js
```diff
@@ -188,18 +133,42 @@ function ImpBookTab({
...
+                            title="Search for a parcel"
+                            onSelect={handleParcelChange}
+                            items={parcelsSearchResults}
+                            loading={parcelsSearchLoading}
+                            fetchItems={searchParcels}
+                            clearSearch={() => clearParcelsSearch}
+                            value={parcelNumber?.toString()}
+                            modal
+                            links={false}
+                            // history={history}
```
lewisrenfrew, line 154: could delete?

src/Service.Host/client/src/containers/importBooks/tabs/ImpBookTab.js
```diff
@@ -56,37 +52,33 @@ const mapStateToProps = state => ({
...
 
 const initialise = props => dispatch => {
     if (
-        !props.employees ||
-        props.employees.length === 0 ||
-        !props.allSuppliers ||
-        props.allSuppliers.length === 0 ||
-        !props.transportCodes ||
+       !props.transportCodes ||
         props.transportCodes.length === 0 ||
```
lewisrenfrew, line 70: these OR's checking whether arrays are null or empty could be rewritten as a single NOT I think, e.g.
<code>
!props.transportCodes?.length
</code>

src/Service/Modules/ParcelsModule.cs
```diff
@@ -44,6 +45,18 @@ private object GetParcels()
                 .WithView("Index");
         }
 
+        private object GetParcelsByNumber()
```
lewisrenfrew, line 48: Could you combine these two GET endpoints and decide what to do based on the values of the binded resource? Seems odd to have 'parcels' and 'parcelsByNumber' as two seperate js itemTypes when they are the same thing, and the only reason I can see for you seperating them is this different GET endpoint

lewisrenfrew (approved):

src/Service.Host/client/src/components/importBooks/ImportBook.js
iaincrawf: D'oh, cheers

src/Service.Host/client/src/components/importBooks/ImpBookPrintOut.js
iaincrawf: It did, but I think for once I actually want it in this file so that there are spaces in the print out

src/Service/Modules/ParcelsModule.cs
iaincrawf: I think I did it because of the action types on the front end - the normal parcels was updateApiActions and parclesbyNumber uses fetchApiActions. The fetchByQueryString in updateApiActions was ideal for the mad search string of multiple things I did in the actual parcels search, but it was going to be a pain to use for this where I just want to be able to call the standard search method without setting up a string like in fetchApiActions. I maybe could use the same resource and method on the backend here, and just check whether it has just the searchTerm or all the other search terms, but then would the front end be smart enough for two sets of actions & selectors for the same url/type? I could try it

src/Service.Host/client/src/components/__tests__/impbooks/Impbook.test.js
iaincrawf: Unfortunately they're values in table cells, so these ones don't have their own labels. I did think about the wrong field having the right value, but reckoned the tests being there in this form were better than not being there and couldn't think of a nice way to get to exactly the right ones. Probably could through some magic selection pain if we really wanted

src/Service.Host/client/src/components/importBooks/tabs/ImpBookTab.js
iaincrawf: I think I forgot to take this out, couldn't get this to work either. I can't remember if I had required on the dropdown as I'm sure that broke it, I'll have a quick play about and see if I can test it

iaincrawf: Yeah, getByLabelText doesn't work for it

PR #244, opened 2021-10-01
Label fix

lewisrenfrew:


iaincrawf (approved):
having to escape all of those "s makes it a bit of a nightmare but yeah lgtm with the quote at the start

lewisrenfrew:
> having to escape all of those "s makes it a bit of a nightmare but yeah lgtm with the quote at the start

Yeah... I should probably write a little function to build the print strings but just trying to get something to work for now

PR #250, opened 2021-10-06
Search for rsns & auto rsn details within order details

iaincrawf:
Two or three more of these to do but thought I'd just put this through on its own so it's a smaller pr


src/Facade/Services/ISuppliersService.cs
```diff
@@ -4,8 +4,8 @@
     using Linn.Common.Facade;
     using Linn.Stores.Domain.LinnApps;
 
-    public interface ISuppliersService
+    public interface IRsnService
```
iaincrawf, line 7: ffs will fix oops

PR #252, opened 2021-10-07
Bonus consignment changes

richardiphillips:


scripts/install.sh
```diff
@@ -7,8 +7,8 @@ dotnet restore
 if [ "$CI" ] && [ "$TRAVIS" ]
 then 
 	source ~/.nvm/nvm.sh; 
-	nvm install 10;
-	nvm use 10;
+	nvm install 14;
+	nvm use 14;
```
iaincrawf, line 11: didn't realise we'd already added these nvm install/uses, nice one

iaincrawf (approved):
Looks good

PR #254, opened 2021-10-08
fix moves request

lewisrenfrew:


lewisrenfrew:
> not sure why this helps but i'm sure it must
 binding nullable int fields was breaking because blank params were toString()ing to ‘null’. A blank string keeps it happy. Might be a nicer way to fix this but this works for me

PR #261, opened 2021-10-14
Stock rotations by part report

lewisrenfrew:
I should probably write some unit tests for the method that gets the report data

### DIFF
--- a/src/Domain.LinnApps/StockLocators/StockLocatorService.cs
+++ b/src/Domain.LinnApps/StockLocators/StockLocatorService.cs
@@ -3,7 +3,6 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
-    using System.Security.Cryptography.X509Certificates;
 
     using Linn.Common.Authorisation;
     using Linn.Common.Persistence;
@@ -350,7 +349,7 @@
                              x.StockPoolCode,
                              x.State,
                              IsPallet = x.PalletNumber.HasValue,
-                             Loc = x.PalletNumber ?? x.StorageLocation.LocationId
+                             Loc = x.PalletNumber ?? x.LocationId
                          }).Select(
                 g => new StockLocator
                          {
--- a/tests/Unit/Domain.LinnApps.Tests/StockLocatorServiceTests/WhenCreatingAndDepartmentCodeEntered.cs
+++ b/tests/Unit/Domain.LinnApps.Tests/StockLocatorServiceTests/WhenCreatingAndDepartmentCodeEntered.cs
@@ -46,6 +46,7 @@
                 .Returns(true);
             this.StoresPalletRepository.FilterBy(Arg.Any<Expression<Func<StoresPallet, bool>>>())
                 .Returns(this.pallets.AsQueryable());
+            
             this.result = this.Sut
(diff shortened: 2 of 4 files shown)


src/Facade/Services/StockQuantitiesService.cs
```diff
@@ -1,35 +1,23 @@
...
-                    .FilterBy(x => x.PartNumber == partNumber.ToUpper());
-                return new SuccessResult<IEnumerable<StockQuantities>>(res);
-            }
-
-            var partNumberPattern = Regex.Escape(partNumber.Trim(' ')).Replace("\\*", ".*?");
-            var r = new Regex(partNumberPattern, RegexOptions.IgnoreCase);
-
-            var result = this.repository
-                .FilterBy(x => r.IsMatch(x.PartNumber));
+            var result = this.repository.FilterByWildcard(partNumber.Replace("*", "%").ToUpper());
```
iaincrawf, line 20: So does the EF.Functions.Like use "%" like oracle as the wildcard character? If so then that's pretty nice, and this'll let users either use % if they're used to that or * if they prefer it

src/Persistence.LinnApps/Repositories/StockLocatorRepository.cs
```diff
@@ -35,6 +35,14 @@ public IQueryable<StockLocator> FilterBy(Expression<Func<StockLocator, bool>> ex
                 .Include(l => l.Part);
         }
 
+        public IQueryable<StockLocator> FilterByWildcard(string search)
```
iaincrawf, line 38: If this works well/is quick, do you think it'd be worth us just making the .Where EF.Functions.Like below our sort of go to implementation of the normal FilterBy, or do you think it's worth keeping it separate? It'll still work fine without a wildcard, won't it?

iaincrawf (approved):
Big fan of the ef "Like" if it's quick, feel like if we used it widely it would make a lot of the old-school oracle users happy

src/Persistence.LinnApps/Repositories/StockLocatorRepository.cs
lewisrenfrew: It is fast, much faster than the query EF evaluates (often clientside in memory) if you pass some regex to achieve 'LIKE' in our FilterBy(...). 

I wouldn't advocate for repalcing FilterBy though. It's nice and open because you can pass in an expression from the outside world that can do anything you like. In theory you could pass an expression with EF.Functions.Like but that would require your calling context to have EntityFramework installed. Often we're calling repository methods from our Domain which certainly doesn't want EntityFramework or whatever as a depdendency. 

I think the solution is to have both in tandem. Maybe write an<code>IQueryable<T> FilterWithWildcard(string searchExpressionWithWildcards)</code>  into our IRepositories.

iaincrawf: Nice! 

Yeah sorry I had a moment and forgot we pass any expression in to the repository filterBy methods and not just a string, both would be good!

richardiphillips: no reason not to have both

PR #266, opened 2021-10-19
Add pva field to impbook so can be used

iaincrawf:
Quick separate pr as folk are wanting to use this field

Added it to the screen and printout, just put it in the first section of fields
image


andrewfraser73 (approved):
looks good

what is the default when creating?

PR #268, opened 2021-10-20
Impb post duty function

iaincrawf:
As mentioned, builds failing due to BuildLinks method in ProcessResult resource builder somehow being called during the integration tests WhenPostingDuty in ImportBooksModuleSpecs

### DIFF
--- a/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
+++ b/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
@@ -7,6 +7,7 @@
     using Linn.Common.Persistence;
     using Linn.Stores.Domain.LinnApps.Exceptions;
     using Linn.Stores.Domain.LinnApps.ExternalServices;
+    using Linn.Stores.Domain.LinnApps.Models;
 
     public class ImportBookService : IImportBookService
     {
@@ -50,7 +51,7 @@
             return exchangeRates;
         }
 
-        public void PostDutyForOrderDetails(
+        public ProcessResult PostDutyForOrderDetails(
             IEnumerable<ImportBookOrderDetail> orderDetails,
             int supplierId,
             int employeeId,
@@ -69,6 +70,8 @@
                     this.transactionManager.Commit();
                 }
             }
+
+            return new ProcessResult(true, "Successfully posted duty");
         }
 
         public void Update(ImportBook from, ImportBook to)
--- a/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
+++ b/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
@@ -4,19 +4,28 @@
     using System.Collections.Generic;
     using System.Linq;
     using Linn.Common.Persistence;
+    using Linn.Stores.Domain.LinnApps.Exceptions;
 
     public class ImportBookService : IImportBookService
     {
         private readonly IRepository<ImportBookExchangeRate, ImportBookExchangeRateKey> exchangeRateRepository;
         private readonly IRepository<LedgerPeriod, int> ledgerPeriodRepository;
         private readonly IRepository<Supplier, int> supplierRepository;
+        private readonly IRepository<ImportBookOrderDetail, ImportBookOrderDetailKey> orderDetailRepository;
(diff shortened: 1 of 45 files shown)


src/Proxy/PurchaseLedgerPack.cs
```diff
@@ -0,0 +1,111 @@
...
+                return int.Parse(res);
+            }
+        }
+
+        public int GetNextLedgerSeq()
+        {
+            using (var connection = this.databaseService.GetConnection())
+            {
+                connection.Open();
+                var cmd = new OracleCommand("cg_code_controls_next_val", connection)
```
iaincrawf, line 46: This is already in the common database service, but I need to get multiple values from it with my domain service (which can't use the common db service) to insert into purchase ledger. Felt it was better to recreate here and just call from domain, than to put extra logic in facade and pass in multiple values

richardiphillips: you can do it in the repository

iaincrawf: Or should I have a purchaseLedgerFacadeService + call add for each detail being posted, and just do the normal databaseService "get next val" call within the createFromResource of the purchaseLedgerFacadeService?

iaincrawf: Ahh will give that a go! Sorry hadn't refreshed didn't see you'd replied

src/Domain.LinnApps/ImportBooks/ImportBookService.cs
```diff
@@ -3,19 +3,84 @@
...
+
+        public ProcessResult PostDutyForOrderDetails(
+            IEnumerable<ImportBookOrderDetail> orderDetails,
+            int supplierId,
+            int employeeId,
+            DateTime postDutyDate)
+        {
+            foreach (var detail in orderDetails)
+            {
+                var oldDetail = this.orderDetailRepository.FindById(
```
richardiphillips, line 67: do any order details have to be passed in here? 
Shouldn't it just deal with what's in the database and disable the button while there are any outstanding changes on the form?

src/Domain.LinnApps/ImportBooks/ImportBookService.cs
```diff
@@ -3,19 +3,84 @@
...
+                {
+                    throw new PostDutyException("Cannot post duty without saving all order details first");
+                }
+
+                if (string.IsNullOrEmpty(oldDetail.PostDuty) && !string.IsNullOrEmpty(detail.PostDuty) && detail.PostDuty.Equals("Y"))
+                {
+                    this.PostDuty(detail, supplierId, employeeId, postDutyDate);
+
+                    oldDetail.PostDuty = "Y";
+                    this.transactionManager.Commit();
```
richardiphillips, line 79: i don't like transactions being committed in the domain (although i did do it once myself and didn't enjoy it and really it should refactor it out). 
If you've got a dedicated facade service method for this anyway might be better to commit it there?

src/Domain.LinnApps/ImportBooks/ImportBookService.cs
```diff
@@ -29,46 +94,53 @@ public void Update(ImportBook from, ImportBook to)
...
-            entity.CancelledReason = to.CancelledReason;
-            entity.NumCartons = to.NumCartons;
-            entity.NumPallets = to.NumPallets;
-            entity.Comments = to.Comments;
-            entity.CreatedBy = to.CreatedBy;
-            entity.CustomsEntryCodePrefix = to.CustomsEntryCodePrefix;
-            entity.Pva = to.Pva;
+            var newPurchaseLedgerEntry = new PurchaseLedger
+                                             {
+                                                 Pltref = pltref,
```
richardiphillips, line 113: as discussed

src/Domain.LinnApps/ImportBooks/ImportBookService.cs
iaincrawf: > the first time Rhona uses this it would be a good idea for you or her to let Leigh know to carefully check the resulting purchase ledger post.

Agreed, will do. I'll wait until Jenkins is back before merging

PR #269, opened 2021-10-20
Part numbers templates fix

lewisrenfrew:
- fix bug where part_number_templates next sequence val was updated  even if the user didn't create the part from a template


andrewfraser73 (approved):
looks ok

PR #280, opened 2021-11-01
Book in rsn

lewisrenfrew:


andrewfraser73 (approved):
Looks ok - the business logic is all hidden away in GOODS_IN_PACK but it should behave properly

PR #281, opened 2021-11-02
Shipfiles no order contact bug fix

lewisrenfrew:
- git diff is making it hard to see whats going on here so: 
- Basically I've moved the loop over all emails-to-be-sent that arrives in the requestResource from the domain to the facade. 
- This is so that I can do transactionManager.Commit()'s on a per email basis so the database gets updated to know which emails sent and which ones didn't in the case of errors. 
- Previously the domain just threw an exception if an email errored and the whole request 500'ed with no data persisting regarding which emails were sent
- see new facade unit tests for clarity

### DIFF
--- a/src/Facade/Services/ConsignmentShipfileFacadeService.cs
+++ b/src/Facade/Services/ConsignmentShipfileFacadeService.cs
@@ -43,9 +43,9 @@
                     try
                     {
                         var sent = this.domainService.SendEmails(
-                        new ConsignmentShipfile { Id = resource.Id, ConsignmentId = resource.ConsignmentId },
-                        toSend.Test,
-                        toSend.TestEmailAddress);
+                            new ConsignmentShipfile { Id = resource.Id, ConsignmentId = resource.ConsignmentId },
+                            toSend.Test,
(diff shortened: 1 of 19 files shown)


src/Facade/Services/ConsignmentShipfileFacadeService.cs
```diff
@@ -35,25 +36,27 @@ public IResult<IEnumerable<ConsignmentShipfile>> GetShipfiles()
...
+                            new ConsignmentShipfile { Id = resource.Id, ConsignmentId = resource.ConsignmentId },
+                            toSend.Test,
+                            toSend.TestEmailAddress);
+                        result.Add(sent);
 
+                        this.transactionManager.Commit();
+                    }
+                    catch (PdfServiceException exception)
+                    {
+                        return new ServerFailureResult<IEnumerable<ConsignmentShipfile>>(exception.Message);
```
iaincrawf, line 55: I'm not sure of context but would it be useful for us to know which email it failed on in the message or would we just have to go and check ourselves anyway?

iaincrawf (approved):
Sounds/looks good

PR #282, opened 2021-11-02
kill parcel ut fields added + front end permission checks

iaincrawf:
Backend already had these fields. Will need to give goods in the admin priv and the orders team 00s the licence to kill before I merge this

Added front end authoriation permission checks, should probs do some back end checks for these too - 
did anyone do backend authorisation checks in stores? Where'd we settle on putting them, facade?

Think I still have impbook backend checks to add too - might do them both on a new branch.


lewisrenfrew (approved):
Best place to do a permission check is the domain... but I doubt I've religiously stuck to that convention myself

PR #285, opened 2021-11-04
parcel & impbook backend auth checks

iaincrawf:
Most just in the module like I was saying, sorry.. Have just used the common FacadeServices so would need to write my own facade services just to pass the privileges through to a new Domain method

### DIFF
--- a/src/Domain.LinnApps/AuthorisedAction.cs
+++ b/src/Domain.LinnApps/AuthorisedAction.cs
@@ -9,5 +9,11 @@
         public const string CreateStockLocator = "stock-locator.create";
 
         public const string UpdateStockLocator = "stock-locator.update";
+
+        public const string ParcelAdmin = "parcel.admin";
+
+        public const string ParcelKillAdmin = "parcel-kill.admin";
+
+        public const string ImpbookAdmin = "import-books.admin";
     }
 }
--- a/src/Domain.LinnApps/ImportBooks/IImportBookService.cs
+++ b/src/Domain.LinnApps/ImportBooks/IImportBookService.cs
@@ -15,6 +15,7 @@
             IEnumerable<ImportBookOrderDetail> orderDetails,
             int supplierId,
             int employeeId,
-            DateTime postDutyDate);
+            DateTime postDutyDate,
+            IEnumerable<string> privileges);
     }
 }
--- a/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
+++ b/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
@@ -12,20 +12,19 @@
 
     public class ImportBookService : IImportBookService
     {
+        private readonly IAuthorisationService authorisationService;
+
         private readonly IRepository<ImportBookExchangeRate, ImportBookExchangeRateKey> exchangeRateRepository;
 
         private readonly IRepository<LedgerPeriod, int> ledgerPeriodRepository;
 
         private readonly IRepository<ImportBookOrderDetail, ImportBookOrderDetailKey> orderDetailRepository;
 
-        private readonly IRepository<PurchaseLedger, int> purchaseLedgerRepository;
-
         private readonly IPurchaseLedgerPack purchaseLedgerPack;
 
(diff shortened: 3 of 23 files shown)


lewisrenfrew (approved):

lewisrenfrew:
I've generally got around passing privileges back to FacadeService<...>'s by sticking them in the request resource.

That way you could do checks in the facade at least if you really don't want to do them in the domain.. Or just leave them in the module I suppose... Just seems likely we are going to rewrite all the modules at some point, in which case it'll be annoying to have domain logic in there

lewisrenfrew:
oops, just accidentally closed the PR... reopened immediately

iaincrawf:
> I've generally got around passing privileges back to FacadeService<...>'s by sticking them in the request resource.
> 
> That way you could do checks in the facade at least if you really don't want to do them in the domain.. Or just leave them in the module I suppose... Just seems likely we are going to rewrite all the modules at some point, in which case it'll be annoying to have domain logic in there

Hmm didn't realise we'd be rewriting them anytime soon, what would that be for? Like if we replace the react front ends years down the line or to get rid of Nancy or something? We have some in production as well and a couple others in modules here so I think I might be bad and leave them for now

richardiphillips:
> > I've generally got around passing privileges back to FacadeService<...>'s by sticking them in the request resource.
> > That way you could do checks in the facade at least if you really don't want to do them in the domain.. Or just leave them in the module I suppose... Just seems likely we are going to rewrite all the modules at some point, in which case it'll be annoying to have domain logic in there
> 
> Hmm didn't realise we'd be rewriting them anytime soon, what would that be for? Like if we replace the react front ends years down the line or to get rid of Nancy or something? We have some in production as well and a couple others in modules here so I think I might be bad and leave them for now

yeah. Definitely don't leave the code in the module.

richardiphillips:
> > > I've generally got around passing privileges back to FacadeService<...>'s by sticking them in the request resource.
> > > That way you could do checks in the facade at least if you really don't want to do them in the domain.. Or just leave them in the module I suppose... Just seems likely we are going to rewrite all the modules at some point, in which case it'll be annoying to have domain logic in there
> > 
> > 
> > Hmm didn't realise we'd be rewriting them anytime soon, what would that be for? Like if we replace the react front ends years down the line or to get rid of Nancy or something? We have some in production as well and a couple others in modules here so I think I might be bad and leave them for now
> 
> yeah. Definitely don't leave the code in the module. I know it doesn't replace checking in the backend but ideally the client should do an auth check and if it doesn't get a create link rel then it shouldn't display the create button so the add backend won't be hit unless we get a hacker.

I would say if anything just leave this code in module for now as it should never be called and make sure the add button or whatever is disabled using the linkrels

richardiphillips:
> > > > > I've generally got around passing privileges back to FacadeService<...>'s by sticking them in the request resource.
> > > > > That way you could do checks in the facade at least if you really don't want to do them in the domain.. Or just leave them in the module I suppose... Just seems likely we are going to rewrite all the modules at some point, in which case it'll be annoying to have domain logic in there
> > > > 
> > > > 
> > > > Hmm didn't realise we'd be rewriting them anytime soon, what would that be for? Like if we replace the react front ends years down the line or to get rid of Nancy or something? We have some in production as well and a couple others in modules here so I think I might be bad and leave them for now
> > > 
> > > 
> > > yeah. Definitely don't leave the code in the module. I know it doesn't replace checking in the backend but ideally the client should do an auth check and if it doesn't get a create link rel then it shouldn't display the create button so the add backend won't be hit unless we get a hacker.
> > 
> > 
> > I would say if anything just leave this code in module for now as it should never be called and make sure the add button or whatever is disabled as per the linkrels
> 
> Leave it or move it? The front end is already disabled if don't have privs, think front end validation is nice for users to see straight away that they can't do something

So you're saying that me saying Move It and Leave It in the same comment was confusing! It was a ramble during which I came to the conclusion that it's fine and  i'd leave it but checking the links to make the client helpful by disabling buttons etc is more important.

richardiphillips (approved):
it's fine. don't listen to any rambling comments that anyone might make at the bottom

PR #291, opened 2021-11-11
Tickets 14822 14808

lewisrenfrew:


andrewfraser73 (approved):
looks ok

src/Domain.LinnApps/Parts/MechPartSourceService.cs
```diff
@@ -63,14 +58,161 @@ public string GetCapacitanceLetterAndNumeralCode(string unit, decimal value)
...
+            current.QualityVerifiedBy = updated.QualityVerifiedBy;
+            current.QualityVerifiedDate = updated.QualityVerifiedDate;
+            current.RemoveTCodeBy = updated.RemoveTCodeBy;
+            current.RemoveTCodeDate = updated.RemoveTCodeDate;
+            current.VerifiedBy = updated.VerifiedBy;
+            current.VerifiedDate = updated.VerifiedDate;
+            current.PurchasingQuotes = updated.PurchasingQuotes;
+            current.Usages = updated.Usages;
+            current.LifeExpectancyPart = updated.LifeExpectancyPart;
+            current.Configuration = updated.Configuration;
```
iaincrawf, line 184: They might already be somewhere but are there tests for these million bits of the MechPartSources to check they update the right thing? Looks fine to my eye but there's just so many

iaincrawf (approved):
I don't really understand any context but code looks good to me anyway 😄

src/Domain.LinnApps/Parts/MechPartSourceService.cs
lewisrenfrew: No.. there are tests for anything thats more complicated than just a this.that = that.that but you'r e right, I should add tests for all of this incredibly boring code

PR #293, opened 2021-11-12
PartTemplate work

RossStewart:
PR for PartTemplate work. Initial backend submitted more submissions and tests to follow.


src/Facade/ResourceBuilders/PartTemplateResourceBuilder.cs
```diff
@@ -15,10 +15,13 @@ public PartTemplateResource Build(PartTemplate model)
...
+                           AllowVariants = model.AllowVariants,
+                           Variants = model.Variants,
                            AccountingCompany = model.AccountingCompany,
                            ProductCode = model.ProductCode,
                            StockControlled = model.StockControlled,
                            LinnProduced = model.LinnProduced,
+                           RmFg = model.RmFg,
                            BomType = model.BomType,
                            AssemblyTechnology = model.AssemblyTechnology,
                            AllowPartCreation = model.AllowPartCreation,
```
richardiphillips, line 27: remember you can set the "self" linkrel in GetLocation down below

src/Persistence.LinnApps/Repositories/PartTemplateRepository.cs
```diff
@@ -35,7 +35,7 @@ public IQueryable<PartTemplate> FindAll()
```
richardiphillips, line 35: remember to take off AsNoTracking up above if we're amending these things now

richardiphillips (approved):
Looks good. Needs a couple of integration tests added for put and post i think.

RossStewart:
No worries, was going to add a few more tests on Monday morning so I’ll make these cha ges then too!

src/Service.Host/client/src/containers/parts/Parts.js
```diff
@@ -6,13 +6,16 @@ import partsSelectors from '../../selectors/partsSelectors';
 import { getPrivileges } from '../../selectors/userSelectors';
 import partTemplatesActions from '../../actions/partTemplatesActions';
 import partTemplatesSelectors from '../../selectors/partTemplatesSelectors';
+import partTemplateActions from '../../actions/partTemplatesActions';
+import partTemplateSelectors from '../../selectors/partTemplatesSelectors';
```
lewisrenfrew, line 11: do we need these extra action creators and selectors in the parts container?

lewisrenfrew: I also just realised I never got you set up witht the linter which would have complained about this most likely. Give me a shout today and I can try and remember what we need to do to get it working

RossStewart: Can talk about it after stand up if you're about

PR #294, opened 2021-11-15
print rsn stuff

lewisrenfrew:


src/Facade/Services/GoodsInFacadeService.cs
```diff
@@ -137,5 +141,12 @@ public IResult<ValidateRsnResult> ValidateRsn(int rsnNumber)
         {
             return new SuccessResult<ValidateRsnResult>(this.domainService.ValidateRsn(rsnNumber));
         }
+
+        public IResult<ProcessResult> PrintRsn(int rsnNumber)
+        {
+            this.printRsnService.PrintRsn(rsnNumber, "Service Copy");
```
richardiphillips, line 147: I'd pass through usernumber from the client too for when you come to use the printer mappings. 
This could end up being used to print rsns for other people at different printers potentially.

richardiphillips (approved):

PR #296, opened 2021-11-18
fix multiple ontos

lewisrenfrew:


src/Domain.LinnApps/GoodsIn/GoodsInService.cs
```diff
@@ -150,7 +150,7 @@ public class GoodsInService : IGoodsInService
                 transactionType,
                 createdBy,
                 partNumber,
-                qty,
+                total,
```
lewisrenfrew, line 153: I think this was the culprit @andrewfraser73. I was only passing the do_bookin routine the qty of one of the lines as opposed to the total. Not sure why this fixes it but it seems to

richardiphillips (approved):

src/Domain.LinnApps/GoodsIn/GoodsInService.cs
andrewfraser73: maybe the qty passed capped off the amount it did but if it sorts it then that goods

andrewfraser73 (approved):

PR #297, opened 2021-11-18
Impbook - allow mismatched weights

iaincrawf:
- Will now let Rhona save if weight mismatch -  if order details total weight doesn't match weight on front tab, save is now enabled but clicking it will bring up a new modal with a big warning that the weights don't match
- Moved cpc numbers to be loaded into the main impbook component instead of order details tab, so that I can print out the display text on the print component
- made page full width

Haven't had a chance to fix the conversion issue, will check the oracle form code and sort that separately tomorrow hopefully

### DIFF
--- a/src/Service.Host/client/src/components/__tests__/impbooks/Impbook.test.js
+++ b/src/Service.Host/client/src/components/__tests__/impbooks/Impbook.test.js
@@ -79,6 +79,61 @@ const item = {
     totalImportValue: 1000
 };
 
+const cpcNumbers = [
+    {
+        cpcNumber: 1,
+        description: '51 00 00'
+    },
+    {
+        cpcNumber: 2,
+        description: '51 00 01'
+    },
+    {
+        cpcNumber: 3,
+        description: '40 00 00'
+    },
+    {
+        cpcNumber: 4,
+        description: '40 00 60'
+    },
+    {
+        cpcNumber: 5,
+        description: '40 00 63'
+    },
+    {
+        cpcNumber: 6,
+        description: 'EC Countries'
+    },
+    {
+        cpcNumber: 7,
+        description: '40 00 09 - SAMPLES ONLY'
+    },
+    {
+        cpcNumber: 8,
+        description: 'Low Value Shipment'
+    },
+    {
+        cpcNumber: 9,
+        description: '40 51 00 - IPR to Free'
+    },
+    {
+        cpcNumber: 10,
+        description: '40 00 58'
+    },
+    {
+        cpcNumber: 11,
+        description: '61 23 F01'
+    },
+    {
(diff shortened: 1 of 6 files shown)


src/Service.Host/client/src/components/__tests__/impbooks/Impbook.test.js
```diff
@@ -116,6 +171,7 @@ describe('On Create', () => {
                 getExchangeRatesForDate={getExchangeRatesForDate}
                 privileges={privileges}
                 loading={false}
+                cpcNumbers={cpcNumbers()}
```
lewisrenfrew, line 174: any reason this is a function and not just a const?

lewisrenfrew: ah, because the componenet expects a function

src/Service.Host/client/src/components/importBooks/ImpBookPrintOut.js
```diff
@@ -60,6 +61,10 @@ function ImpBookPrintOut({
         return date ? moment(date).format('DD-MMM-YYYY') : '-';
```
lewisrenfrew, line 61: would probably rename to toDdMMYYYY(date) or something

src/Service.Host/client/src/components/importBooks/ImportBook.js
```diff
@@ -354,11 +348,24 @@ function ImportBook({
...
-            `${calcRemainingDuty()}` !== '0' ||
-            `${calcRemainingWeight()}` !== '0'
+            `${calcRemainingDuty()}` !== '0'
         );
     };
 
+    const [dialogOpen, setDialogOpen] = useState(false);
+
+    const handleSaveClick = () => {
+        if (`${calcRemainingWeight()}` !== '0' && !dialogOpen) {
```
lewisrenfrew, line 358: why not calcRemainingWeight() !== 0
i.e. why toString both sides of the expression?

lewisrenfrew (approved):

src/Service.Host/client/src/components/importBooks/ImpBookPrintOut.js
iaincrawf: Yeah I'll rename that haha can probably do better than getNicerDate

src/Service.Host/client/src/components/importBooks/ImportBook.js
iaincrawf: Good question, I'll try it and see. I think I did it for safety cause of js comparisons

iaincrawf:
I'll sort these in the next one, Rhona is wanting to use the old form so going to put this through to stop her

PR #299, opened 2021-11-19
Addrsn

andrewfraser73:
Adds an Add RSN button to the consignment button which if you select RSN will fill out weights, dims and descriptions from the RSN.

Weights and dims come from extension to export rsns view.


lewisrenfrew (approved):

src/Facade/Services/ExportRsnService.cs
```diff
@@ -14,16 +15,27 @@ public ExportRsnService(IQueryRepository<ExportRsn> repository)
             this.repository = repository;
         }
 
-        public IResult<IEnumerable<ExportRsn>> SearchRsns(int accountId, int? outletNumber)
+        public IEnumerable<ExportRsn> FindMatchingRSNs(IEnumerable<ExportRsn> rsns, string searchTerm)
+        {
+            if (string.IsNullOrEmpty(searchTerm))
+            {
+                return rsns;
+            }
```
lewisrenfrew, line 23: stylecop tells me to put new lines after closing of an if block

PR #305, opened 2021-11-25
Tpk unallocate buttons

lewisrenfrew:


tests/Unit/Domain.LinnApps.Tests/TpkServiceTests/WhenUnpickingStock.cs
```diff
@@ -0,0 +1,58 @@
...
+        [SetUp]
+        public void SetUp()
+        {
+            this.ReqMovesRepository.FilterBy(Arg.Any<Expression<Func<ReqMove, bool>>>())
+                .Returns(this.moves.AsQueryable());
+
+            this.StoresPack.UnpickStock(1, 1, Arg.Any<int>(), 1, 1, 1m, 1, 1)
+                .Returns(new ProcessResult(true, string.Empty));
+
+            this.result = this.Sut.UnpickStock(1, 1, 1, 1, 1, 1, 1);
```
iaincrawf, line 37: could maybe use different numbers just to check the right thing's passed in the right place

iaincrawf (approved):
Looks good, was this something from oracle forms? It couldn't be used for this http://wiki.linn.co.uk/doku.php/support#stock_allocation_not_cleared_on_cancelled_req could it?

tests/Unit/Domain.LinnApps.Tests/TpkServiceTests/WhenUnpickingStock.cs
lewisrenfrew: Its just buttons that let them remove one item or all items from a consignment from the TPK screen. 
For some reason I had assumed that the buttons opened a different form and had missed them out.
Not sure it'd be much use anywhere else

PR #317, opened 2021-12-09
ipr report - only show 8 digits of tariff codes

iaincrawf:
"Can I ask if you could amend the report to only show 8 digit tariff nos instead of 10? Customs had said it only needs 8 and I have to delete the last two digits.

Thanks
Yvonne"

http://rndsupport.linn.co.uk/scp/tickets.php?id=14986


iaincrawf:
Sorry ran my code formatter on the domain report service so it's all switched about, really the only changes are the new FormatTariffCodeToEightDigits method and using that in the report and export methods

lewisrenfrew (approved):

PR #320, opened 2021-12-15
Stores parts UI tests

RossStewart:
Write front end tests for the Part Template utility/review code


src/Service.Host/client/src/components/__tests__/parts/PartTemplate.test.js
```diff
@@ -0,0 +1,128 @@
...
+                updateItem={updateItem}
+                setEditStatus={setEditStatus}
+                privileges={['part.admin']}
+                itemId={1}
+            />
+        );
+    });
+
+    test('page renders all fields without crashing...', () => {
+        expect(screen.queryByText('PartRoot')).toBeNull();
```
lewisrenfrew, line 58: What are you testing here?  The test name (renders without crashin) doesn't seem to match the test which is testing whether something is null

src/Service.Host/client/src/components/__tests__/parts/PartTemplate.test.js
```diff
@@ -0,0 +1,128 @@
...
+                updateItem={updateItem}
+                setEditStatus={setEditStatus}
+                item={item}
+                privileges={['part.admin']}
+            />
+        );
+    });
+
+    test('Page renders populated fields', () => {
+        expect(screen.getAllByText('Part Root')[0]).toBeInTheDocument();
```
lewisrenfrew, line 89: why getAllByText? Should there be more than one matching element

src/Service.Host/client/src/components/__tests__/parts/PartTemplateRender.test.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import '@testing-library/jest-dom/extend-expect';
+import { cleanup, screen } from '@testing-library/react';
+import render from '../../../test-utils';
+import PartTemplateSearch from '../../parts/PartTemplatesSearch';
```
lewisrenfrew, line 5: convention is to name the test file the same as the componenet. So this file should be called PartTemplateSearch.test.js. 

Although I think in this case i'd rename your component since it isn't technically a search

src/Service.Host/client/src/components/__tests__/parts/PartTemplateRender.test.js
```diff
@@ -0,0 +1,52 @@
...
+afterEach(() => cleanup());
+
+describe('When creating...', () => {
+    beforeEach(() => defaultRender({ editStatus: 'create' }));
+
+    test('Should Render Title', () => {
+        expect(screen.getByText('Part Template Utility')).toBeInTheDocument();
+    });
+
+    test('Should render Create button', () => {
```
lewisrenfrew, line 45: could do a test to check create button isnt rendered if no /create links in resource

src/Service.Host/client/src/components/__tests__/parts/PartTemplate.test.js
```diff
@@ -0,0 +1,128 @@
...
+        );
+    });
+    test('cannot edit fields', () => {
+        expect(screen.getByDisplayValue('REACT')).toBeDisabled();
+        expect(screen.getByDisplayValue('123')).toBeDisabled();
+        expect(screen.getByDisplayValue('MECHANICAL')).toBeDisabled();
+        expect(screen.getByDisplayValue('A test')).toBeDisabled();
+        expect(screen.getByDisplayValue('Integration UNIT')).toBeDisabled();
+    });
+});
```
lewisrenfrew, line 128: could do some editing/create tests for completeness (i.e. check update/add function mocks get called with correct args)

src/Service.Host/client/src/components/parts/PartTemplate.js
```diff
@@ -0,0 +1,520 @@
...
+        setPartTemplate(x => ({ ...x, [propertyName]: newValue }));
+    };
+
+    const handleProductCodeChange = product => {
+        handleFieldChange('productCode', product.name);
+        handleFieldChange('productAnalysisCodeDescription', product.description);
+    };
+
+    const allowedToEdit = () => {
+        return privileges?.some(priv => priv === 'part.admin');
```
lewisrenfrew, line 140: I thought we were replacing these front end privilege checks with Hypermedia links

src/Service.Host/client/src/components/parts/PartTemplatesSearch.js
```diff
@@ -0,0 +1,86 @@
...
+function PartTemplateSearch({ privileges, partTemplates }) {
+    const [template, setTemplate] = useState();
+
+    const createUrl = () => {
+        return `/inventory/part-templates/create`;
+    };
+    const viewUrl = () => {
+        return `/inventory/part-templates/${template}`;
+    };
+    const canCreate = () => {
```
lewisrenfrew, line 17: same as before. You have the presence or absence of a '/create' link in your resource now. Might as well use it.

src/Service.Host/client/src/containers/parts/PartTemplates.js
```diff
@@ -0,0 +1,26 @@
+import { connect } from 'react-redux';
+import { initialiseOnMount } from '@linn-it/linn-form-components-library';
+import partTemplatesActions from '../../actions/partTemplatesActions';
+import partTemplatesSelectors from '../../selectors/partTemplatesSelectors';
+import partTemplateSelectors from '../../selectors/partTemplateSelectors';
+import PartTemplatesSearch from '../../components/parts/PartTemplatesSearch';
+import { getPrivileges } from '../../selectors/userSelectors';
```
lewisrenfrew, line 7: you won't need to pass privileges into any of your components if you use the resources links I wouldn't think

tests/Integration/Service.Tests/PartsModuleSpecs/WhenGettingPartTemplates.cs
```diff
@@ -9,6 +9,8 @@
     using Linn.Stores.Domain.LinnApps.Parts;
     using Linn.Stores.Resources.Parts;
 
+    using Microsoft.EntityFrameworkCore.Internal;
```
lewisrenfrew, line 12: looks suspect

src/Service.Host/client/src/components/__tests__/parts/PartTemplateRender.test.js
lewisrenfrew: I realised that you aren't using the presence or absence of a /create link to decide whether user can create as I read further down the PR, so above comment doesn't make sense now

lewisrenfrew (approved):
Looks good. 

I would change your components to use the hypermedia links as opposed to explicit privilege checks so that you reap the rewards of all your response model efforts.

Could do with a few more front end tests too, specifically to test what data actually gets passed to the update and save actions. I think that's probably the most important logic to test  as it concerns what actual data gets back to the back end. 

But if you are understandably sick of react tests then feel free to leave em

RossStewart:
Yeah, definitely thought about the Update tests - will definitely add them now that I have a bit of a better understanding on how it's working. Will make the suggested changes and think about the update ones, would be good to have them.

src/Facade/ResourceBuilders/PartTemplateResourceBuilder.cs
```diff
@@ -1,45 +1,67 @@
...
-            return $"/inventory/part-templates/{model.PartRoot}";
+            return $"/inventory/part-templates/{model.ResponseData.PartRoot}";
         }
 
-        private IEnumerable<LinkResource> BuildLinks(PartTemplate model)
+        private IEnumerable<LinkResource> BuildLinks(ResponseModel<PartTemplate> model)
         {
+            if (this.authorisationService.HasPermissionFor(AuthorisedAction.PartAdmin, model.Privileges))
+            {
+                yield return new LinkResource { Rel = "create", Href = this.GetLocation(model) };
```
lewisrenfrew, line 61: create href should point to your /create endpoint

PR #321, opened 2021-12-16
Create Consignment Typeaheads

andrewfraser73:
Create Consignment typeaheads
- sales account id lookup
- not one but two address lookups one just any address id and one to get it from the outlet default address


src/Service/Modules/SalesOutletModule.cs
```diff
@@ -27,14 +28,26 @@ private object GetSalesOutlets()
...
                 .WithModel(result)
                 .WithMediaRangeModel("text/html", ApplicationSettings.Get())
                 .WithView("Index");
         }
+
+        private object GetSalesOutletAddresses()
+        {
+            var resource = this.Bind<SalesOutletRequestResource>();
+
+            IResult<IEnumerable<SalesOutlet>> result = this.salesOutletService.GetOutletAddresses(resource.accountId, resource.SearchTerm);
```
lewisrenfrew, line 45: is there an integration for this new route?

lewisrenfrew (approved):

PR #324, opened 2021-12-21
add period number back onto impbook

iaincrawf:
get period number and add it in on create and on update if date created changes, tests updated to check for this

### DIFF
--- a/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
+++ b/src/Domain.LinnApps/ImportBooks/ImportBookService.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Reflection.PortableExecutable;
 
     using Linn.Common.Authorisation;
     using Linn.Common.Persistence;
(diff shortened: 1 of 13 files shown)


src/Domain.LinnApps/ImportBooks/ImportBookService.cs
```diff
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Reflection.PortableExecutable;
```
lewisrenfrew, line 6: seems suspect

tests/Unit/Domain.LinnApps.Tests/ImportBookServiceTests/WhenAddingInvoiceDetail.cs
```diff
@@ -3,11 +3,16 @@
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Linq.Expressions;
+
+    using Castle.DynamicProxy.Generators.Emitters.SimpleAST;
```
lewisrenfrew, line 8: suspect

lewisrenfrew (approved):

andrewfraser73 (approved):
looks ok for getting period on there

tests/Unit/Domain.LinnApps.Tests/ImportBookServiceTests/WhenAddingInvoiceDetail.cs
iaincrawf: good spots, were sus indeed

PR #330, opened 2022-01-13
move wwd pack calls into one connection / db session

iaincrawf:
hopefully will stop wwd report from failing half of the time as will hopefully make the two calls go to the same session

### DIFF
--- a/src/Domain.LinnApps/ExternalServices/IWWDPack.cs
+++ b/src/Domain.LinnApps/ExternalServices/IWWDPack.cs
@@ -2,8 +2,6 @@
 {
     public interface IWwdPack
     {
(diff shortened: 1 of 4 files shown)


lewisrenfrew (commented):
did it make any difference locally? hard to say i guess cos of how intermittently it fails. worth trying to see if you can break it locally before merging anyway

lewisrenfrew (approved):

iaincrawf:
> did it make any difference locally? hard to say i guess cos of how intermittently it fails. worth trying to see if you can break it locally before merging anyway

I think it seems to be fixed, been refreshing things and starting and stopping which I think was breaking it yesterday and hasn't broken at all.. fingers crossed for live

PR #335, opened 2022-01-25
Link parts ut to part suppliers

lewisrenfrew:


src/Facade/ResourceBuilders/PartResourceBuilder.cs
```diff
@@ -135,6 +135,24 @@ private IEnumerable<LinkResource> BuildLinks(Part part)
...
                                  Href = $"/inventory/stock-locators?partId={part.Id}"
                              };
+
+            if (part.PreferredSupplierId.HasValue && part.PreferredSupplierId != 4415)
+            {
+                yield return new LinkResource
+                                 {
+                                     Rel = "part-supplier",
+                                     Href =
+                                         $"/purchasing/part-suppliers/record?partId={part.Id}&supplierId={part.PreferredSupplierId}"
```
richardiphillips, line 145: what's the /record for?

richardiphillips (approved):
nice link rel usage

src/Facade/ResourceBuilders/PartResourceBuilder.cs
lewisrenfrew: tl;dr: Basically to keep the breadcrumbs sensible.

I put /record in on the purchasing side since  /part-suppliers is already in use for the search page.

since  the Part Supplier Id consistes of two fields...

Another option would have been /purchasing/part-suppliers/<partId>/<supplierId>
Which is kind of semantically correct since a single part can have multiple Part Suppliers.

But it still messes up the breadcrumbs in their current form

 I preferred /record?<ids as query params>

PR #337, opened 2022-01-28
Conditionally access current user

lewisrenfrew:
- I have no reason to believe this'll make any difference but worth a try


RossStewart (approved):

richardiphillips (approved):
do we need the requiresauth for the get?

lewisrenfrew:
> do we need the requiresauth for the get?

Probably not, but just trying this for now. 

https://app.linn.co.uk/inventory/part-templates is broken but works fine on sys and locally so I'm a bit stumped.

I'll remove the .requiresAuthentication() if this works

PR #343, opened 2022-02-11
Quantity int to decimal conversion on StockLocators

RossStewart:
Wasn't entirely sure if I needed to add the WWD work ones here but can review that.

### DIFF
--- a/src/Domain.LinnApps/StockLocators/StockLocator.cs
+++ b/src/Domain.LinnApps/StockLocators/StockLocator.cs
@@ -10,7 +10,7 @@
     {
         public int Id { get; set; }
 
-        public int? Quantity { get; set; }
+        public decimal? Quantity { get; set; }
 
         public int? PalletNumber { get; set; }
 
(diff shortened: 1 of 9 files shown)


lewisrenfrew:
> Wasn't entirely sure if I needed to add the WWD work ones here but can review that.

not sure about the wwd ones

RossStewart:
> > > > Wasn't entirely sure if I needed to add the WWD work ones here but can review that.
> > > 
> > > 
> > > not sure about the wwd ones
> > 
> > 
> > Yeah, wasn't too sure myself. I can see that they're being used in some domain test (WhenGettingWhatWillDecrementReport) but wasn't sure if this would cause a problem for Mark. Really I guess we'd need to know the function of the WWD_WORK_DETAILS and WWD_WORK tables.
> 
> It probably doesn't really matter. You could check the table columns to see if they're integers or not, or just leave them as ints since it hasn't caused any problems so far.

Table has them defined as datatype NUMBER so it looks like it's possible decimal places could be a thing? I think I'll just revert them to ints for now as we know that that's working and if Mark sees an future issue we can always change them.

PR #345, opened 2022-02-15
HasExportReturn

andrewfraser73:
Export_RSNS_view now serves both making an export return and adding an RSN to a consignment

there is a new flag on the view to show if a RSN is already in an export return.  If it is it shouldn't be allowed to be added to another export return but it can go on a consignment.  If it isn't it should be allowed on an export return but can't go on a consignment.

There is also a small change to add carrier ref to consignments

### DIFF
--- a/tests/Unit/Facade.Tests/ExportRsnService/WhenGettingExportRsnsWithNumber.cs
+++ b/tests/Unit/Facade.Tests/ExportRsnService/WhenGettingExportRsnsWithNumber.cs
@@ -25,7 +25,7 @@
 
             this.ExportRsnRepository.FilterBy(Arg.Any<Expression<Func<ExportRsn, bool>>>()).Returns(rsns.AsQueryable());
 
-            this.result = this.Sut.SearchRsns(123, null, "2","Y");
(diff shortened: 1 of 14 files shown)


tests/Unit/Facade.Tests/ExportRsnService/WhenGettingExportRsnsWithNumber.cs
```diff
@@ -25,7 +25,7 @@ public void SetUp()
 
             this.ExportRsnRepository.FilterBy(Arg.Any<Expression<Func<ExportRsn, bool>>>()).Returns(rsns.AsQueryable());
 
-            this.result = this.Sut.SearchRsns(123, null, "2");
+            this.result = this.Sut.SearchRsns(123, null, "2","Y");
```
richardiphillips, line 28: looks like your stylecop might have switched off if it's not complaining about the space here.

richardiphillips (approved):

lewisrenfrew (approved):

iaincrawf (approved):

PR #352, opened 2022-02-21
Allow sending of previous shipfile emails

lewisrenfrew:


tests/Integration/Service.Tests/WandModuleSpecs/WhenAddingShipfile.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        [Test]
+        public void ShouldCallFacadeService()
+        {
+            this.ShipfileService.Received().Add(Arg.Any<ConsignmentShipfileResource>());
+        }
+
+        [Test]
+        public void ShouldReturnDeleted()
```
richardiphillips, line 60: deleted?

src/Facade/Services/ConsignmentShipfileFacadeService.cs
```diff
@@ -71,5 +76,31 @@ public IResult<ConsignmentShipfile> DeleteShipfile(int id)
...
+
+            if (consignment == null)
+            {
+                return new BadRequestResult<ConsignmentShipfile>("Consignment Not Found");
+            }
+
+            var shipfile = new ConsignmentShipfile
+                               {
+                                   ConsignmentId = consignment.ConsignmentId,
+                                   Id = this.repository.FindAll().Max(x => x.Id) + 1
```
iaincrawf, line 98: do we need to specify an id for shipfiles? I haven't seen us have to do this ^ before, just seen us setting an Id for ones that use the get_next_val function or whatever (doesn't mean we don't need to here though)

iaincrawf (approved):

src/Facade/Services/ConsignmentShipfileFacadeService.cs
richardiphillips, line 98: i think this should be done in the repo

tests/Unit/Facade.Tests/ConsignmentShipfilesFacadeServiceTests/WhenAdding.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class WhenAdding: ContextBase
+    {
+        private IResult<ConsignmentShipfile> result;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.ConsignmentRepository.FindBy(Arg.Any<Expression<Func<Consignment, bool>>>())
+                .Returns(new Consignment {});
+            this.Repository.FindAll().Returns(new List<ConsignmentShipfile> { new ConsignmentShipfile { Id = 1 } }.AsQueryable());
```
richardiphillips, line 28: is this findall used?

richardiphillips (approved):

src/Facade/Services/ConsignmentShipfileFacadeService.cs
richardiphillips: with a sequence would be better

tests/Unit/Facade.Tests/ConsignmentShipfilesFacadeServiceTests/WhenAdding.cs
lewisrenfrew: It is for generating the Id, but you're previous comment to just use a sequence is a better idea anyway

src/Facade/Services/ConsignmentShipfileFacadeService.cs
lewisrenfrew: Using a sequence is a better idea anyway

PR #355, opened 2022-03-03
Consignment Export Edit

andrewfraser73:
A consignment export info editing page.  Current consignment page doesn't allow for editing after the consignment closed and there is another page in linnapps that this will replace to add carrier references and customs entry details.

For EU consignments all the shipments that go out on the day get the same master carrier ref so added functionality to set this for a bunch of clicked consignments as they were moaning about going in and out of screens.

This is a first draft and may change when I show it to them and they think of other amendments.

### DIFF
--- a/src/Facade/Services/ConsignmentFacadeService.cs
+++ b/src/Facade/Services/ConsignmentFacadeService.cs
@@ -227,7 +227,6 @@
                 var fromDate = DateTime.Parse(resource.From);
                 var toDate = DateTime.Parse(resource.To);
 
-
                 if (resource.HubId == null)
                 {
                     return new SuccessResult<IEnumerable<Consignment>>(this.repository.FilterBy(c => c.DateClosed >= fromDate && c.DateClosed <= toDate));
--- a/src/Facade/Services/ConsignmentFacadeService.cs
+++ b/src/Facade/Services/ConsignmentFacadeService.cs
@@ -11,12 +11,15 @@
     using Linn.Stores.Domain.LinnApps.Consignments;
     using Linn.Stores.Domain.LinnApps.Exceptions;
     using Linn.Stores.Resources.Consignments;
+    using Linn.Stores.Resources.RequestResources;
 
-    public class ConsignmentFacadeService : FacadeService<Consignment, int, ConsignmentResource, ConsignmentUpdateResource>
+    public class ConsignmentFacadeService : FacadeService<Consignment, int, ConsignmentResource, ConsignmentUpdateResource>, IConsignmentFacadeService
     {
         private readonly IConsignmentService consignmentService;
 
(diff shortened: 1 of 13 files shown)


src/Facade/Services/IConsignmentFacadeService.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    using Linn.Common.Facade;
+    using Linn.Stores.Domain.LinnApps.Consignments;
+    using Linn.Stores.Resources.Consignments;
+    using Linn.Stores.Resources.RequestResources;
+    using System.Collections.Generic;
+
+    public interface IConsignmentFacadeService : IFacadeService<Consignment, int, ConsignmentResource, ConsignmentUpdateResource>
+    {
+        IResult<IEnumerable<Consignment>> GetByRequestResource(ConsignmentsRequestResource resource);
```
lewisrenfrew, line 11: is this extra method just to let you do filtering on a resource? If so then IFacadeFilterService gets you that

src/Facade/Services/ConsignmentFacadeService.cs
```diff
@@ -62,21 +66,31 @@ protected override void UpdateFromResource(Consignment entity, ConsignmentUpdate
...
             if (updateResource.Status == "C")
             {
-                if (entity.Status != "L")
+                if (entity.Status == "L")
                 {
-                    throw new ConsignmentEditException($"You cannot edit closed consignment {entity.ConsignmentId}.");
-                }
+                    if (!updateResource.ClosedById.HasValue)
+                    {
+                        throw new ConsignmentCloseException("Closed by id must be provided to close consignment");
```
lewisrenfrew, line 73: I like to do this sort of stuff in a domainService these days

src/Facade/Services/ConsignmentFacadeService.cs
```diff
@@ -200,5 +214,31 @@ private void UpdateItems(Consignment entity, IList<ConsignmentItemResource> upda
...
+            {
+                if (string.IsNullOrEmpty(resource.To))
+                {
+                    return new BadRequestResult<IEnumerable<Consignment>>("No To Date specified");
+                }
+
+                var fromDate = DateTime.Parse(resource.From);
+                var toDate = DateTime.Parse(resource.To);
+
+
```
lewisrenfrew, line 230: blanky

lewisrenfrew (approved):

src/Facade/Services/IConsignmentFacadeService.cs
andrewfraser73: there is a bit of logic in there can't see from other uses of IFacadeFilterService how it would help me but might be missing something

PR #362, opened 2022-04-07
alter nominal/dept search

iaincrawf:
Played about with this after Chris mentioned IT department example at standup.

Now does exact match search first to bring back things like "IT" for department, + fixed typo/copy paste so that also searches on department.description and not department.departmentCode twice


RossStewart (approved):
Looks good!

PR #369, opened 2022-04-26
Stock loc search additions

lewisrenfrew:


src/Proxy/StockLocatorLocationsViewService.cs
```diff
@@ -39,6 +41,22 @@ public StockLocatorLocationsViewService(IDatabaseService databaseService)
                         : $"PART_NUMBER = '{partNumber.ToUpper()}'");
             }
 
+            if (partDescription != null)
```
richardiphillips, line 44: sting.isnullorempty?

richardiphillips (approved):

src/Proxy/StockLocatorLocationsViewService.cs
lewisrenfrew: yes!

PR #395, opened 2022-08-29
stop certain changes on amends

andrewfraser73:
stop being able to change decrement rule or bom type unless creating


lewisrenfrew (approved):
You could also stop these fields changing at the domain level and enshrine those rules with tests if you were feeling extra passionate about locking them down

PR #414, opened 2022-10-12
Limit nominals

lewisrenfrew:


andrewfraser73 (approved):
approved can't see a reason to need a nominal account in stores without stores posts allowed

PR #434, opened 2022-12-21
Wwd local report

richardiphillips:
Local version of Report Table with ability to display text colour attributes. 
Will add to real report table once it's working


lewisrenfrew (approved):

lewisrenfrew:
- do you think there's any point trying to squeeze this into a DataGrid when you come to  add it into the real report table? I guess there's not a huge benefit but it does give nice features like sorting, filtering, searching etc

richardiphillips:
> * do you think there's any point trying to squeeze this into a DataGrid when you come to  add it into the real report table? I guess there's not a huge benefit but it does give nice features like sorting, filtering, searching etc

The front end change to ReportTable is pretty tiny so will do that to begin with but I would love to try datagrid. I'll promote it up my list of things to be done

PR #435, opened 2022-12-22
Label Printer

lewisrenfrew:
- this has been on my lists for a while
- basically just provides a url to the QcLabelPrintScreen pop up on the goods in utility
- allows reprinting labels to a different printer etc
- will hopefully close http://rndsupport.linn.co.uk/scp/tickets.php?id=16704


richardiphillips:
apart from js tests obviously :-)

PR #441, opened 2023-02-08
Bom Freq on Part Record

RossStewart:


src/Persistence.LinnApps/ServiceDbContext.cs
```diff
@@ -465,9 +465,9 @@ private void BuildParts(ModelBuilder builder)
             e.Property(p => p.SingleSourcePart).HasColumnName("SINGLE_SOURCE_PART").HasMaxLength(1);
             e.Property(p => p.SafetyCertificateExpirationDate).HasColumnName("SAFETY_CERTIFICATE_EXPIRY_DATE");
             e.Property(p => p.SafetyDataDirectory).HasColumnName("SAFETY_DATA_DIRECTORY").HasMaxLength(500);
-            e.Property(p => p.LinnProduced).HasColumnName("LINN_PRODUCED")
-                .HasMaxLength(1).HasColumnType("VARCHAR2").IsUnicode(false);
-            e.Property(p => p.BomType).HasColumnName("BOM_TYPE").HasMaxLength(1).HasColumnType("VARCHAR2").IsUnicode(false);
+            e.Property(p => p.LinnProduced).HasColumnName("LINN_PRODUCED").HasMaxLength(1);
+            e.Property(p => p.BomType).HasColumnName("BOM_TYPE").HasMaxLength(1);
+            e.Property(p => p.BomVerifyFreqWeeks).HasColumnName("BOM_VERIFY_FREQ_WEEKS").HasMaxLength(4);
```
lewisrenfrew, line 470: I take it those .IsUnicode(false) calls you deleted were redundant? I don't remember why they were there in the first place

lewisrenfrew (approved):

src/Persistence.LinnApps/ServiceDbContext.cs
RossStewart: Will be resubmitting this with my changes in.  Had a bit of a branch issue but I'll explain it in stand up.

lewisrenfrew: what's the issue? All looks ok to me apart from the test fail which is an easy fix

lewisrenfrew: Why does that matter? you merged master into this branch anyway and the diff looks sensible - I would have just fixed the test here and merged this branch (unless I'm missing something)

richardiphillips: was it telling you that it couldn't be merged? Seems like all you need to do is fix the test like Lewis says

RossStewart: Reverted the removal of the Unicode (false) calls too. Wasn't intentional. That was partly why I wasn't convinced I was totally up to date there but it's fine.

PR #445, opened 2023-04-26
BK

andrewfraser73:
BahnhofKoenig (Station Master) is the app for adding warehouse tasks
This is a web based version

It relies on expanded WCS_PACK for much of the functionality 
The idea when the new AGV system arrives will be to rewrite just Warehouse_Service to proxy out to new routes instead of calling WCS_PACK

### DIFF
--- a/src/Facade/Services/WarehouseFacadeService.cs
+++ b/src/Facade/Services/WarehouseFacadeService.cs
@@ -1,16 +1,22 @@
 ﻿namespace Linn.Stores.Facade.Services
 {
     using Linn.Common.Facade;
+    using Linn.Common.Persistence;
     using Linn.Stores.Domain.LinnApps;
     using Linn.Stores.Domain.LinnApps.Models;
+    using Linn.Stores.Domain.LinnApps.Wcs;
+    using Linn.Stores.Resources;
 
     public class WarehouseFacadeService : IWarehouseFacadeService
     {
         private readonly IWarehouseService warehouseService;
 
-        public WarehouseFacadeService(IWarehouseService warehouseService)
+        private readonly IRepository<Employee, int> employeeRepository;
+
+        public WarehouseFacadeService(IWarehouseService warehouseService, IRepository<Employee, int> employeeRepository)
         {
             this.warehouseService = warehouseService;
+            this.employeeRepository = employeeRepository;
         }
 
         public IResult<MessageResult> MoveAllPalletsToUpper()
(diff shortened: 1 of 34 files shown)


src/Proxy/WcsPack.cs
```diff
@@ -60,21 +58,301 @@ public void MovePalletToUpper(int palletNumber, string reference)
...
+                        Value = taskSource
+                });
+
+            cmd.Parameters.Add(
+                new OracleParameter("p_who", OracleDbType.Int32)
+                    {
+                        Direction = ParameterDirection.Input,
+                        Value = who
+                    });
+
```
lewisrenfrew, line 216: blanky blank blank

src/Facade/Services/WarehouseFacadeService.cs
```diff
@@ -23,5 +29,82 @@ public IResult<MessageResult> MovePalletToUpper(int palletNumber, string referen
...
+        public IResult<WarehouseLocation> GetPalletAtLocation(string location)
+        {
+            var warehouseLocation = this.warehouseService.GetWarehouseLocation(location, null);
+            if (warehouseLocation?.PalletId == null)
+            {
+                return new NotFoundResult<WarehouseLocation>("Pallet not on system");
+            }
+
+            return new SuccessResult<WarehouseLocation>(warehouseLocation);
+        }
```
lewisrenfrew, line 108: some of this feels a bit domain logic-y to me... although I'm not familair with the domain at all

lewisrenfrew (approved):

src/Facade/Services/WarehouseFacadeService.cs
andrewfraser73: Nearly all the critical domain logic is down in the WCS_PACK and whatever replaces it.  There is checking of some parameters to determine if it is a valid request such as whether is it likely to be a valid priority or pallet number which does slightly feel a bit domain-y.  In reality the WCS_PACK or the call to the SCS PLC will come back with an error if you submit an invalid parameter so this is a bit of pre-optimization to not submit something to them that you know won't ever work.   So it might be stripped out at some point when this is enhanced to not merely say a success/fail on submitting a task but more detailed explanation for failure.

src/Facade/Services/WarehouseFacadeService.cs
```diff
@@ -23,5 +29,82 @@ public IResult<MessageResult> MovePalletToUpper(int palletNumber, string referen
...
         }
+
+        public IResult<MessageResult> MakeWarehouseTask(WarehouseTaskResource resource)
+        {
+            if (resource == null)
+            {
+                return new BadRequestResult<MessageResult>("No task to add");
+            }
+
+            if (resource.PalletNumber < 1 || resource.PalletNumber > 4000)
```
richardiphillips, line 40: I agree with @lewisrenfrew 
This is definitely domain code

PR #446, opened 2023-05-01
Stock Triggers w/ Integration 

RossStewart:


src/Facade/ResourceBuilders/StockTriggerLevelResourceBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+                           MaxCapacity = stockTriggerLevel.MaxCapacity,
+                           PalletNumber = stockTriggerLevel.PalletNumber,
+                           KanbanSize = stockTriggerLevel.KanbanSize,
+                           Links = this.BuildLinks(stockTriggerLevel).ToArray()
+                       };
+        }
+
+        public string GetLocation(StockTriggerLevel stockTriggerLevel)
+        {
+            return $"/inventory/stock-trigger-levels/{stockTriggerLevel.LocationId}";
```
lewisrenfrew, line 29: should this not be based on the Id field as opposed to the LocationId field

src/Facade/Services/StockTriggerLevelsFacadeService.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var toDelete = this.repository.FindBy(s => s.Id == id);
+            this.repository.Remove(toDelete);
+            return new SuccessResult<StockTriggerLevel>(toDelete);
+        }
+
+        protected override StockTriggerLevel CreateFromResource(StockTriggerLevelsResource resource)
+        {
+            var newStockTriggerLevel = new StockTriggerLevel
+                               {
```
lewisrenfrew, line 35: How is the Id field generated?

src/Persistence.LinnApps/Repositories/StockTriggerLevelsRepository.cs
```diff
@@ -33,5 +33,20 @@ public IQueryable<StockTriggerLevel> FindAll()
...
+
+        public void Add(StockTriggerLevel entity)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void Remove(StockTriggerLevel entity)
+        {
+            throw new NotImplementedException();
+        }
```
lewisrenfrew, line 50: don't you need to implement all these methods?

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,98 @@
...
+            return this.Negotiate.WithModel(result).WithMediaRangeModel("text/html", ApplicationSettings.Get);
+        }
+
+        private object UpdateStockTriggerLevel(int id)
+        {
+            if (!this.authorisationService.HasPermissionFor(
+                    AuthorisedAction.UpdateStockTriggerLevel,
+                    this.Context.CurrentUser.GetPrivileges()))
+            {
+                return new UnauthorisedResult<StockTriggerLevel>(
```
lewisrenfrew, line 77: I'd be tempted to write some tests that cover these auth checks if they are important

lewisrenfrew (approved):

src/Facade/ResourceBuilders/StockTriggerLevelResourceBuilder.cs
```diff
@@ -0,0 +1,39 @@
...
+    using Linn.Stores.Domain.LinnApps.StockLocators;
+    using Linn.Stores.Resources;
+
+    public class StockTriggerLevelResourceBuilder : IResourceBuilder<StockTriggerLevel>
+    {
+        public StockTriggerLevelsResource Build(StockTriggerLevel stockTriggerLevel)
+        {
+            return new StockTriggerLevelsResource
+            {
+                           PartNumber = stockTriggerLevel.PartNumber,
```
richardiphillips, line 17: i'd put the id in here

src/Facade/ResourceBuilders/StockTriggerLevelResourceBuilder.cs
richardiphillips, line 29: in fact you definitely need the id as this isn't right.

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,98 @@
+﻿namespace Linn.Stores.Service.Modules
+{
+    using System.Collections.Generic;
```
richardiphillips, line 3: a few unused imports here

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,98 @@
...
+            this.Post("/inventory/stock-trigger-levels/create", _ => this.CreateStockTriggerLevel());
+            this.Get("/inventory/stock-trigger-levels/{id:int}", parameters => this.GetStockTriggerLevel(parameters.id));
+            this.Get("/inventory/stock-trigger-levels", _ => this.SearchStockTriggerLevels());
+            this.Put("/inventory/stock-trigger-levels/{id:int}", parameters => this.UpdateStockTriggerLevel(parameters.id));
+            this.Delete("/inventory/stock-trigger-levels/{id:int}", parameters => this.DeleteStockTriggerLevel(parameters.id));
+        }
+
+        private object SearchStockTriggerLevels()
+        {
+            var resource = this.Bind<SearchRequestResource>();
```
richardiphillips, line 46: this method seems untested

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,98 @@
...
+        private object GetStockTriggerLevel(int id)
+        {
+            var results = this.stockTriggerLevelsFacadeService.GetById(id);
+            return this.Negotiate.WithModel(results).WithMediaRangeModel("text/html", ApplicationSettings.Get)
+                .WithView("Index");
+        }
+
+        private object DeleteStockTriggerLevel(int id)
+        {
+            return this.Negotiate.WithModel(this.stockTriggerLevelsFacadeService.DeleteStockTriggerLevel(id));
```
richardiphillips, line 95: i'd definitely check for create authorisation here at least

src/Facade/Services/StockTriggerLevelsFacadeService.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public StockTriggerLevelsFacadeService(IRepository<StockTriggerLevel, int> repository, ITransactionManager transactionManager) : base(repository, transactionManager)
+        {
+            this.repository = repository;
+            this.transactionManager = transactionManager;
+        }
+
+        public IResult<StockTriggerLevel> DeleteStockTriggerLevel(int id)
+        {
+            var toDelete = this.repository.FindBy(s => s.Id == id);
```
richardiphillips, line 27: i would at least log that this has happened. maybe capture the user number too. 
I was going to say that you don't need the transaction manager variable above but you probably need it here as I don't think this is being committed in the module either

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/ContextBase.cs
```diff
@@ -0,0 +1,65 @@
+﻿namespace Linn.Stores.Service.Tests.StockTriggerLevelsModuleSpecs
+{
+    using System.Collections.Generic;
+    using System.Security.Claims;
+
+    using Linn.Common.Authorisation;
+    using Linn.Common.Facade;
+    using Linn.Common.Persistence;
```
richardiphillips, line 8: no persistence stuff is being used here

tests/Unit/Domain.LinnApps.Tests/StockLocatorServiceTests/ContextBase.cs
```diff
@@ -41,7 +41,7 @@ public class ContextBase
 
         protected IRepository<ReqMove, ReqMoveKey> ReqMoveRepository { get; private set; }
 
-        protected IQueryRepository<StockTriggerLevel> triggerLevelRepository { get; private set; }
+        protected IRepository<StockTriggerLevel, int> TriggerLevelRepository { get; private set; }
```
richardiphillips, line 46: blank blank

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenGettingById.cs
```diff
@@ -0,0 +1,54 @@
...
+    using NUnit.Framework;
+
+    public class WhenGettingById : ContextBase
+    {
+        private StockTriggerLevel stockTriggerLevel1;
+
+        [SetUp]
+        public void SetUp()
+        {
+            this.stockTriggerLevel1 = new StockTriggerLevel { PalletNumber = 123, PartNumber = "SAMPLE PART", TriggerLevel = 1, LocationId = 35, KanbanSize = 1, MaxCapacity = 1 };
```
richardiphillips, line 22: pernickity but it's nicer to shorten these long lines for readability

src/Persistence.LinnApps/Repositories/StockTriggerLevelsRepository.cs
```diff
@@ -33,5 +33,20 @@ public IQueryable<StockTriggerLevel> FindAll()
         {
             throw new NotImplementedException();
         }
+
+        public StockTriggerLevel FindById(int key)
+        {
+            throw new NotImplementedException();
```
richardiphillips, line 39: Does this work when you run it? These are all being used by the facade service so will need to be implemented I expect.

richardiphillips (approved):
just a bit of tidying needed

src/Facade/Services/StockTriggerLevelsFacadeService.cs
```diff
@@ -0,0 +1,92 @@
...
+                                           };
+
+            return newStockTriggerLevel;
+        }
+
+        protected override void UpdateFromResource(
+            StockTriggerLevel entity,
+            StockTriggerLevelsResource updateResource)
+        {
+            entity.Id = updateResource.Id;
```
lewisrenfrew, line 78: probably shouldn't allow for the possibility of the Id updating since its a primary key

src/Facade/Services/StockTriggerLevelsFacadeService.cs
```diff
@@ -0,0 +1,92 @@
...
+            this.transactionManager.Commit();
+
+            return new SuccessResult<StockTriggerLevel>(toDelete);
+        }
+
+        protected override StockTriggerLevel CreateFromResource(StockTriggerLevelsResource resource)
+        {
+            var newStockTriggerLevel = new StockTriggerLevel 
+                                           {
+                                               Id = this.databaseService.GetNextVal("STL_ID"),
```
lewisrenfrew, line 62: I think GetNextVal looks in CG_CODE_CONTROLS for a sequence. I don't see a CG_CODE_CONTROLS entry fro STL_ID. Does your INSERT work like this? If it does then there must be a trigger or something generating an Id or you, and this line can be removed

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,103 @@
...
+                .WithView("Index");
+        }
+
+        private object CreateStockTriggerLevel()
+        {
+            if (!this.authorisationService.HasPermissionFor(
+                    AuthorisedAction.CreateStockTriggerLevel,
+                    this.Context.CurrentUser.GetPrivileges()))
+            {
+                return new UnauthorisedResult<StockTriggerLevel>(
```
lewisrenfrew, line 55: Your tests are looking for a BadRequest (400) not an Unauthorised (401), so I'm not sure how they're passing. My guess would be the framework doesn't know what to do with UnauthorisedResult<StockTriggerLevel>'s, and so just BadRequest's by default. I think I'd change these to return BadRequest instead of Unauthorised just to be safe

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,103 @@
...
+            return this.Negotiate.WithModel(result).WithMediaRangeModel("text/html", ApplicationSettings.Get);
+        }
+
+        private object UpdateStockTriggerLevel(int id)
+        {
+            if (!this.authorisationService.HasPermissionFor(
+                    AuthorisedAction.UpdateStockTriggerLevel,
+                    this.Context.CurrentUser.GetPrivileges()))
+            {
+                return new UnauthorisedResult<StockTriggerLevel>(
```
lewisrenfrew, line 72: BadRequest again

src/Service/Modules/StockTriggerLevelsModule.cs
```diff
@@ -0,0 +1,103 @@
...
+                .WithView("Index");
+        }
+
+        private object DeleteStockTriggerLevel(int id)
+        {
+            if (!this.authorisationService.HasPermissionFor(
+                    AuthorisedAction.CreateStockTriggerLevel,
+                    this.Context.CurrentUser.GetPrivileges()))
+            {
+                return new UnauthorisedResult<StockTriggerLevel>(
```
lewisrenfrew, line 94: and again

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenAdding.cs
```diff
@@ -0,0 +1,81 @@
...
+                                            PartNumber = "PART",
+                                            Id = 1,
+                                            KanbanSize = 1,
+                                            LocationId = 1,
+                                            MaxCapacity = 1,
+                                            PalletNumber = 1,
+                                            TriggerLevel = 1
+                                        };
+
+            this.AuthorisationService.HasPermissionFor("stock-trigger-level.create", Arg.Any<IEnumerable<string>>()).Returns(true);
```
lewisrenfrew, line 45: I'd use your AuthorisedAction constants in here as opposed to strings, just to prevent future typos

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenDeleting.cs
```diff
@@ -0,0 +1,55 @@
...
+    using NSubstitute;
+
+    using NUnit.Framework;
+
+    public class WhenDeleting : ContextBase
+    {
+        [SetUp]
+        public void SetUp()
+        {
+            this.AuthorisationService.HasPermissionFor("stock-trigger-level.create", Arg.Any<IEnumerable<string>>()).Returns(true);
```
lewisrenfrew, line 22: use AuthorisdeAction const again

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenUpdating.cs
```diff
@@ -0,0 +1,83 @@
...
+                                            PartNumber = "PART",
+                                            Id = 1,
+                                            KanbanSize = 1,
+                                            LocationId = 1,
+                                            MaxCapacity = 1,
+                                            PalletNumber = 1,
+                                            TriggerLevel = 1
+                                        };
+
+            this.AuthorisationService.HasPermissionFor("stock-trigger-level.update", Arg.Any<IEnumerable<string>>()).Returns(true);
```
lewisrenfrew, line 47: AuthorisedAction const again

lewisrenfrew (approved):

PR #456, opened 2023-06-19
Stock Trigger Feedback

RossStewart:
Expand search + add some typeahead fixes


tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenSearchingPalletNumbers.cs
```diff
@@ -16,7 +16,7 @@
 
     using NUnit.Framework;
 
-    public class WhenSearching : ContextBase
+    public class WhenSearchingPalletNumbers : ContextBase
```
lewisrenfrew, line 19: this test looks functionally identical to the one above.. whats the need for both?

lewisrenfrew (approved):

tests/Integration/Service.Tests/StockTriggerLevelsModuleSpecs/WhenSearchingPalletNumbers.cs
lewisrenfrew: both tests are just binding 'storagePlaceSearchTerm' from the request and everything after that point is mocked out anyway, so i think you can just delete one of them

RossStewart: Yeah, fair point. Will do.

PR #460, opened 2023-07-18
all the bits

lewisrenfrew:
- allows steven w and co to add parts (via a source sheet) with altium info, and update said info of existing parts (via a new tab on the parts ut)
- <s>still need to update the plsql to make the new info propogate over when creating a part from a source sheet</s> done


andrewfraser73 (approved):
looks ok

PR #466, opened 2023-10-20
catch possibility of user running stoc loc when not logged in 

lewisrenfrew:
(although I kind of think this is impossible)


lewisrenfrew:
http://rndsupport.linn.co.uk/scp/tickets.php?id=17050

richardiphillips (approved):

PR #467, opened 2023-11-14
Make SCS Backup from existing warehouse pallets

andrewfraser73:
This is the logic that makes JSON of existing WCS pallets with rotationaverage etc.

to make it work you need to put it into rabbit scs.x exchange with routing key scs.store.pallets

### DIFF
--- a/src/Domain.LinnApps/Wcs/WarehousePallet.cs
+++ b/src/Domain.LinnApps/Wcs/WarehousePallet.cs
@@ -1,0 +1,50 @@
+﻿namespace Linn.Stores.Domain.LinnApps.Wcs
+{
+    using System;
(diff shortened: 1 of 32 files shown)


src/Domain.LinnApps/Wcs/WarehousePallet.cs
```diff
@@ -0,0 +1,50 @@
...
+                case "L":
+                    return 0;
+                default:
+                    return 1;
+            }
+        }
+
+        public int ScsHeat()
+        {
+            // SpeedFactor is too unrelia
```
lewisrenfrew, line 37: ble

lewisrenfrew (approved):

PR #470, opened 2023-12-15
Fix Delete /display names STL

RossStewart:


tests/Unit/Facade.Tests/StockTriggerLevelsServiceTests/WhenAdding.cs
```diff
@@ -0,0 +1,69 @@
...
+    using Linn.Stores.Domain.LinnApps.StockLocators;
+    using Linn.Stores.Resources;
+
+    using NSubstitute;
+
+    using NUnit.Framework;
+
+    public class WhenAdding : ContextBase
+    {
+        private readonly int stockTriggerLevelId = 1;
```
lewisrenfrew, line 15: var looks unused

tests/Unit/Facade.Tests/StockTriggerLevelsServiceTests/WhenAdding.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        private StockTriggerLevelsResource resource;
+
+        private IResult<StockTriggerLevel> result;
+
+        [SetUp]
+        public void SetUp()
+        {
+            var stockTriggerLevel = new StockTriggerLevel
```
lewisrenfrew, line 24: is this used?

tests/Unit/Facade.Tests/StockTriggerLevelsServiceTests/WhenAdding.cs
```diff
@@ -0,0 +1,69 @@
...
+                TriggerLevel = 400
+            };
+
+            this.result = this.Sut.Add(this.resource);
+        }
+
+        [Test]
+        public void ShouldAdd()
+        {
+            this.Repository.Received().Add(Arg.Any<StockTriggerLevel>());
```
lewisrenfrew, line 52: What are we actually testing here? the fact that the repository is called and that the Created varaint of the IResult are returned are just functions of the common library code - you shouldn't be unit testing them here. If you want to test the /create route an integration test would be a better place to do so

RossStewart: Fair point. It's a fairly standard Add so I suppose I could take this out.

lewisrenfrew (approved):

PR #475, opened 2024-01-24
Storepallets

andrewfraser73:
This is the endpoint that lets you take the SCS JSON Store Pallets format and post it to 
http://localhost:51698/logistics/wcs/warehouse-pallets/scs

which makes 2000-ish basic pallet records in SCS_PALLETS

The EF version did all the fields but took 20 minutes to run - the fast awful code "Fast" repository can do it around 5s but only writes the essential fields we would need back in Oracle pallet number, location and height.  

This will enable us to fix Despatch Pallet Queue with just a SQL tweak to V_DPS_LINN on go live.  It also gives us a pretty good lifeboat if we need to go back as the data is already in Oracle with the info we'd need to recreate a map.  Finally we could also tweak some of the older WCS reports to look at this info and they'd still essentially work.

### DIFF
--- a/src/Domain/ScsPallet.cs
+++ b/src/Domain/ScsPallet.cs
@@ -1,7 +1,5 @@
 ﻿namespace Linn.Stores.Domain
 {
-    using Linn.Stores.Domain.LinnApps.Scs;
-
     public class ScsPallet : StoresAddress
     {
         public int PalletNumber { get; set; }
--- a/src/Domain/ScsPallet.cs
+++ b/src/Domain/ScsPallet.cs
@@ -1,4 +1,4 @@
-﻿namespace Linn.Stores.Domain
+﻿namespace Linn.Stores.Domain.LinnApps.Scs
 {
     public class ScsPallet : StoresAddress
     {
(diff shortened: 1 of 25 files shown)


src/Domain/ScsPallet.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Linn.Stores.Domain
+﻿namespace Linn.Stores.Domain.LinnApps.Scs
```
lewisrenfrew, line 1: should namespace match file location?

src/Domain/StoresAddress.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Linn.Stores.Domain
+﻿namespace Linn.Stores.Domain.LinnApps.Scs
```
lewisrenfrew, line 1: same here

lewisrenfrew (approved):

src/Domain/ScsPallet.cs
andrewfraser73: files got moved about a bit - thought I'd caught them all will fix

PR #478, opened 2024-03-05
Ticket18609

andrewfraser73:
Very basic addition of three filters to move log for location, trans type and stock pool.

this is probably not final version but to show Stores guys so they can say if they like layout and if they want typeaheads of searches for some of those filters.

### DIFF
--- a/src/Domain.LinnApps/Reports/StoresMoveLogReportService.cs
+++ b/src/Domain.LinnApps/Reports/StoresMoveLogReportService.cs
@@ -6,7 +6,6 @@
 
     using Linn.Common.Persistence;
     using Linn.Common.Reporting.Models;
-    using Linn.Stores.Domain.LinnApps.Workstation;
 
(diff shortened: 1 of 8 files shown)


src/Domain.LinnApps/Reports/StoresMoveLogReportService.cs
```diff
@@ -24,23 +25,75 @@ public ResultsModel GetReport(DateTime fromDate, DateTime toDate, string partNum
...
+            }
+
+            return title;
+        }
+
+        private bool MatchesFilter(StoresMoveLog moveLog, string transType, string location, string stockPool)
+        {
+            if (!string.IsNullOrEmpty(location))
+            {
+                if (!moveLog.MatchesLocation(location))
```
lewisrenfrew, line 72: and operator cleaner than nested ifs?

lewisrenfrew (approved):

PR #483, opened 2024-03-22
Consignmentfrominvoice

andrewfraser73:
added two bits of functionality
- lookup of consignment from invoice number on consignments page
- ability to add additional information on address labels


src/Facade/Services/ConsignmentFacadeService.cs
```diff
@@ -239,5 +244,28 @@ public IResult<IEnumerable<Consignment>> GetByRequestResource(ConsignmentsReques
 
             return this.GetAll();
         }
+
+        public IResult<IEnumerable<Consignment>> GetByInvoiceNumber(int invoiceNumber)
+        {
+            var invoice = this.invoiceRepository.FindById(invoiceNumber);
+            if (invoice == null)
+            {
+                return new SuccessResult<IEnumerable<Consignment>>(new List<Consignment>());
```
richardiphillips, line 253: not found?

src/Facade/Services/ConsignmentFacadeService.cs
```diff
@@ -239,5 +244,28 @@ public IResult<IEnumerable<Consignment>> GetByRequestResource(ConsignmentsReques
...
+            }
+
+            if (invoice.ConsignmentId == null)
+            {
+                return new SuccessResult<IEnumerable<Consignment>>(new List<Consignment>());
+            }
+
+            var consignment = this.repository.FindById(invoice.ConsignmentId.Value);
```
richardiphillips, line 261: should move all of this logic down to the domain service.

richardiphillips (approved):

src/Facade/Services/ConsignmentFacadeService.cs
andrewfraser73: was originally notfound but it caused problems in the front end with the loading spinner.  if it had been a direct lookup of saying consignments/123 that returned nothing then i'd be notfounding but its more like the query returns no resullts which makes the front end work better

PR #487, opened 2024-04-18
Add altium data to parts

lewisrenfrew:
- Basically adds a bunch of new fields to the parts utility, which will trickle through to the altium 'libraries' for all the different part types
- you might be wodnering if this still would be better off living on (or indeed already lives on) the source sheets, which would be the case in an ideal world. But loads of the parts steven wants to appear in Altium don't have source sheets, and I felt this was less work than trying to retrospectively create hundreds of source sheets for parts that have already been 'sourced'
- This mini project has actually led me towards questioning the point in even having source sheets at all... something to discuss with Andrew when he's back


richardiphillips:
> * he point in even having source sheets at all... something to discuss with Andrew when he's back

was me that wrote that back in the day. They have merged a bit over the years but original usage was that R&D can use this to collate info before a part has to be made and in theory not all source sheets will become parts as they experiment with things that never make it to production

PR #505, opened 2024-08-23
Jacki Mechanical Sourcing Sheet 

RossStewart:


src/Domain.LinnApps/Parts/PartService.cs
```diff
@@ -345,17 +345,35 @@ public Part CreateFromSource(int sourceId, int createdBy, IEnumerable<PartDataSh
...
+                    "Parts Utility",
+                    $"New Source Sheet Created - {source.PartNumber}",
+                    $"Click here to view: https://app.linn.co.uk/parts/sources/{source.Id}",
+                    null,
+                    null);
+            }
+            else
+            {
+                this.emailService.SendEmail(
+                    "mechanicalsourcingsheet@linn.co.uk",
```
lewisrenfrew, line 365: Is this a real email address / inbox that makes it through to real people? if so avoid leaking it into code (especially in public repos like this one) since spam bots will be scouring the web for them. Could stick this in a config.env or something?

src/Service.Host/client/src/components/parts/mechPartSource/tabs/ProposalTab.js
```diff
@@ -277,9 +291,8 @@ function ProposalTab({
                 <Dropdown
                     label="Assembly Type*"
                     propertyName="assemblyType"
-                    items={['SM', 'TH']}
+                    items={['TH', 'SM']}
```
lewisrenfrew, line 294: why reorder / how come the order of these matters?

src/Service.Host/client/src/components/parts/mechPartSource/tabs/ProposalTab.js
```diff
@@ -104,14 +119,23 @@ function ProposalTab({
...
-                <Grid item xs={3} />
+                <Grid item xs={3}>
+                    <Dropdown
+                        label="Part Type*"
+                        propertyName="partType"
+                        items={mechanicalPartTypes}
+                        fullWidth
+                        value={partType}
+                        onChange={handleFieldChange}
+                    />
```
lewisrenfrew, line 137: would probably be better to just to vary  the items list according to partType === 'E' as opposed to whole Dropdown component

src/Service.Host/client/src/components/parts/mechPartSource/tabs/ProposalTab.js
```diff
@@ -182,17 +206,7 @@ function ProposalTab({
...
-                <DatePicker
-                    label="Production Date"
-                    value={productionDate}
-                    onChange={value => {
-                        handleFieldChange('productionDate', value);
-                    }}
-                />
-            </Grid>
-            <Grid item xs={8} />
+            <Grid item xs={12} />
```
lewisrenfrew, line 209: Don't think this does anything?

src/Domain.LinnApps/Parts/PartService.cs
```diff
@@ -345,17 +345,35 @@ public Part CreateFromSource(int sourceId, int createdBy, IEnumerable<PartDataSh
...
+                    cc,
+                    null,
+                    "stores@linn.co.uk",
+                    "Parts Utility",
+                    $"New Source Sheet Created - {source.PartNumber}",
+                    $"Click here to view: https://app.linn.co.uk/parts/sources/{source.Id}",
+                    null,
+                    null);
+            }
+            else
```
lewisrenfrew, line 362: i would do else if source.MechnicalOrElectrical == "M" here just to be specific. I guess in theory its still possible to create a source sheet that is neither Electrical or Mechanical... although can't really think why that would ever happen

src/Service.Host/client/src/components/parts/mechPartSource/tabs/ProposalTab.js
```diff
@@ -299,6 +312,7 @@ function ProposalTab({
                     fetchItems={searchDepartments}
                     links={false}
                     clearSearch={clearSearchDepartments}
+                    required
```
lewisrenfrew, line 315: I dunno if this flag really does anything? If projectCode is becoming a mandatory field when creating source sheets, a condition/check in the domain code would be better

tests/Unit/Domain.LinnApps.Tests/PartServiceTests/WhenCreatingMechanicalPartFromSourceSheet.cs
```diff
@@ -0,0 +1,48 @@
...
+    using NUnit.Framework;
+
+    public class WhenCreatingMechanicalPartFromSourceSheet : ContextBase
+    {
+        [SetUp]
+        public void SetUp()
+        {
+            this.PhoneList.FindBy(Arg.Any<Expression<Func<PhoneListEntry, bool>>>()).Returns(
+                new PhoneListEntry { EmailAddress = "user1@linn.co.uk", User = new AuthUser { Name = "user1" } });
+
```
lewisrenfrew, line 20: can probaly drop this mock returns

lewisrenfrew (approved):

src/Domain.LinnApps/Parts/PartService.cs
lewisrenfrew: it also might be worth finding out who has access to this inbox if it is a real invoice? If its the same group of people that I've hardcoded on lines 331-333 then there's maybe no need for the if/else at all? If the recipients are the same then the if/else branches are functionally identical

PR #506, opened 2024-08-27
Updated Yml

RossStewart:


richardiphillips (approved):

lewisrenfrew (approved):
what are the test addresses used for? if its just for local testing then we dont need to add them to amazon

lewisrenfrew:
> > what are the test addresses used for? if its just for local testing then we dont need to add them to amazon
> 
> Have them in the tests, was mostly just keeping it consistent with SHIP FILES which also has a test entry. If you pull main on stores you should see the other part of this. Happy to ditch them if they're surplus!

Yeah, remove from them fron any deploy scripts etc if they are only used for tests. Actually I don't think even tests running locally have access to config.env? I could be wrong but I'd remove any references to ConfigurationManger from tests anyway - just use Arg.Any<string> or something? tests shouldn't depend on environment variables really

PR #564, opened 2026-04-15
Print Rsn in new world - test PR

RossStewart:


RossStewart:
> looks ok but might be worth hanging onto the message resources if this might go back to json

Will reinstate them. Good to try it and can come back for a refactor.

